PR #181, opened 2024-12-12
Feature/mc 438 run tool call inference with batches on tool

mc-dorzo:


src/parlant/adapters/nlp/openai.py
```diff
@@ -363,7 +363,7 @@ def __init__(
     async def get_schematic_generator(self, t: type[T]) -> OpenAISchematicGenerator[T]:
         if t == ToolCallInferenceSchema:
             return FallbackSchematicGenerator(
-                GPT_4o_Mini[t](self._logger),  # type: ignore
+                # GPT_4o_Mini[t](self._logger),  # type: ignore
```
kichanyurd, line 366: Don't leave this commented out; remove it. Remove `Fallback` in general here, just return `GPT_4o`.

src/parlant/core/engines/alpha/engine.py
```diff
@@ -356,7 +356,7 @@ async def _do_process(
             ):
                 message_generation_inspections.append(
                     MessageGenerationInspection(
-                        generation=event_generation_result.generation_info,
+                        generation=event_generation_result.generations[0],
```
kichanyurd, line 359: It's incorrect to do `generations[0]` here. What if there were no generations? What if there was more than one? This is coincidental programming at this point.

What really happened here was that `EventGenerationResult` took on a different contract with respect to tool calls and messages. So, instead of forcing them both still under one heading, maybe we don't need to have a reusable dataclass here anymore.

Remove `event_generation.py`, add `MessageEventGenerationResult` under the message generation module, and `ToolEventGenerationResult` under the tool call generation module.

PR #183, opened 2024-12-14
Feature/mc 463 add cachedschematicgenerator

mc-dorzo:


src/parlant/adapters/vector_db/transient.py
```diff
@@ -0,0 +1,301 @@
...
+        self._embedder_factory = embedder_factory
+        self._embedder_type = embedder_type
+
+        self._multi_tenant: nano_vectordb.MultiTenantNanoVDB
+        self._collection_name_to_tenant_id: dict[str, str] = {}
+        self._collections: dict[str, _TransientCollection[BaseDocument]] = {}
+
+    async def __aenter__(self) -> Self:
+        self._embedder = self._embedder_factory.create_embedder(self._embedder_type)
+        self._multi_tenant = nano_vectordb.MultiTenantNanoVDB(self._embedder.dimensions)
```
kichanyurd, line 56: Why do we need the multi tenant version here?

src/parlant/adapters/vector_db/transient.py
```diff
@@ -0,0 +1,301 @@
...
+    async def __aexit__(
+        self,
+        exc_type: Optional[type[BaseException]],
+        exc_value: Optional[BaseException],
+        traceback: Optional[object],
+    ) -> None:
+        pass
+
+    @override
+    async def create_collection(
```
kichanyurd, line 68: You're returning a private class. It's a design bug to expose a private class. Either return `VectorCollection` or make `TransientVectorCollection` public (no underscore prefix).

tests/test_utilities.py
```diff
@@ -35,10 +52,34 @@
...
+    Session,
+    SessionId,
+    SessionStore,
+)
 from parlant.core.tools import LocalToolService, ToolId, ToolResult
+from parlant.core.persistence.common import ObjectId
+from parlant.core.persistence.document_database import DocumentCollection
 
 T = TypeVar("T")
+GLOBAL_CACHE_FILE = Path("_schematic_cache.json")
```
kichanyurd, line 77: Not standard form to name the file with an underscore prefix. Files aren't python attributes :)
Let's call it `schematic_generation_cache.json`

tests/test_utilities.py
```diff
@@ -35,10 +52,34 @@
...
+CACHE_SCHEMATIC_GENERATION_RESULTS = (
+    True
+    if os.environ.get("CACHE_SCHEMATIC_GENERATION_RESULTS", "True").lower() == "true"
```
kichanyurd, line 80: Why do we need this?

tests/test_utilities.py
```diff
@@ -312,3 +353,150 @@ def get_when_done_or_timeout(
         sleep(1)
 
     raise TimeoutError()
+
+
+TBaseModel = TypeVar("TBaseModel", bound=DefaultBaseModel)
+
+
+class _SchematicGenerationResultDocument(TypedDict, total=False):
```
kichanyurd, line 361: Not sure why we need the whole shabang with `DocumentDatabase` and models here. Why not just `json.dump/load` to a dict?

mc-dorzo: Instead start to manage inside the transient/flush all over again (I don't want every call for generation to start json.dump/load, its just takes more time, and with this implementation its more clean and easier. I actually started with json.dump/load and moved to this once I understand that I dont want to load it everytime..

PR #184, opened 2024-12-14
Feature/mc 453 add version command to parlant server cli

mc-dorzo:


src/parlant/bin/server.py
```diff
@@ -484,6 +484,11 @@ def main() -> None:
         default="info",
         help="Log level",
     )
+    @click.option(
+        "--version",
+        is_flag=True,
+        help="Show the version and exit",
```
kichanyurd, line 490: `help="Print server version and exit"`

src/parlant/bin/server.py
```diff
@@ -496,7 +501,12 @@ def cli(
         cerebras: bool,
         together: bool,
         log_level: str,
+        version: bool,
     ) -> None:
+        if version:
+            print(f"parlant-server version {VERSION}")
```
kichanyurd, line 507: Just `print(VERSION)`

kichanyurd: Or actually, `f"Parlant v{VERSION}"`

PR #185, opened 2024-12-14
Feature/mc 464 add modules

mc-dorzo:


src/parlant/bin/server.py
```diff
@@ -133,6 +134,7 @@ class CLIParams:
     port: int
     nlp_service: str
     log_level: str
+    module: str
```
kichanyurd, line 137: Should be plural `modules: list[str]`

src/parlant/bin/server.py
```diff
@@ -183,6 +185,41 @@ async def create_agent_if_absent(agent_store: AgentStore) -> None:
         await agent_store.create_agent(name=DEFAULT_AGENT_NAME)
 
 
+@asynccontextmanager
+async def load_module(
+    container: Container,
+    module_path: str,
+) -> AsyncIterator[None]:
+    mod = importlib.import_module(module_path)
```
kichanyurd, line 193: `mod`? Why not `module`?

src/parlant/bin/server.py
```diff
@@ -183,6 +185,41 @@ async def create_agent_if_absent(agent_store: AgentStore) -> None:
...
+
+    if module_path.endswith("parlant_config"):
```
kichanyurd, line 195: config shouldn't be a module but a `.toml` file. I.e.
```toml
[parlant]
modules = [ "module_1", "module_2" ]
```

src/parlant/bin/server.py
```diff
@@ -369,6 +406,9 @@ async def load_app(params: CLIParams) -> AsyncIterator[ASGIApplication]:
             evaluator=container[BehavioralChangeEvaluator],
         )
 
+        if params.module:
```
kichanyurd, line 409: `if params.modules` ... `load_modules(...)`

plural

kichanyurd: But isn't there a bug here? Like even if the param isn't specified but the modules come from the config file, you should load them.

src/parlant/modules/examples/parlant_bank.py
```diff
@@ -0,0 +1,46 @@
+from lagom import Container
```
kichanyurd, line 1: Good you have examples but they shouldn't go under `src/parlant/modules/` but rather directly under `examples/modules` in the repo root dir.

src/parlant/modules/examples/parlant_bank.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+    await _background_task_service.start(
+        server.serve(),
+        tag="plugin-server 'parlant-bank'",
+    )
+
+    server_instance = server
+
+    service_registry = container[ServiceRegistry]
+    await service_registry.update_tool_service(
```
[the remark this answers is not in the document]
mc-dorzo: So I created as you asked with PluginServer which means local_transient considered as "sdk" by the services api. so when list_services will also include the tools inside. is that fine?

src/parlant/bin/server.py
```diff
@@ -182,6 +186,49 @@ async def create_agent_if_absent(agent_store: AgentStore) -> None:
...
 
 
+@asynccontextmanager
+async def load_modules(
+    container: Container,
+    modules: list[str],
+) -> AsyncIterator[None]:
+    if not modules:
+        config_file = "parlant_config.toml"
+        if os.path.isfile(config_file):
```
kichanyurd, line 196: Don't work with strings and `os.path` but with `Path` objects

src/parlant/bin/server.py
```diff
@@ -182,6 +186,49 @@ async def create_agent_if_absent(agent_store: AgentStore) -> None:
...
+            config = toml.load(config_file)
+            # Expecting structure of:
+            # [parlant]
+            # modules = ["module_1", "module_2"]
+            modules = config.get("parlant", {}).get("modules", [])
+        else:
+            modules = []
+
+    if not modules:
+        container[Logger].info("no modules loaded")
```
kichanyurd, line 206: `LOGGER.info("No modules selected")`

src/parlant/bin/server.py
```diff
@@ -182,6 +186,49 @@ async def create_agent_if_absent(agent_store: AgentStore) -> None:
...
+        yield
+        return
+
+    imported_modules = []
+
+    for module_path in modules:
+        module = importlib.import_module(module_path)
+        if not hasattr(module, "initialize_module") or not hasattr(module, "shutdown_module"):
+            raise StartupError(
+                f"Module {module.__name__} must define initialize_module and shutdown_module"
```
kichanyurd, line 216: `"...initialize_module(container: lagom.Container) and shutdown_module()"`

kichanyurd: And `f"Module '{module.__name__}' ..."`

src/parlant/bin/server.py
```diff
@@ -182,6 +186,49 @@ async def create_agent_if_absent(agent_store: AgentStore) -> None:
...
+            )
+        imported_modules.append(module)
+
+    for m in imported_modules:
+        container[Logger].info(f"initialize module {m.__name__}")
```
kichanyurd, line 221: `LOGGER.info(f"Initializing module '{m.__name__}'")`

src/parlant/bin/server.py
```diff
@@ -182,6 +186,49 @@ async def create_agent_if_absent(agent_store: AgentStore) -> None:
...
+        await m.initialize_module(container)
+
+    try:
+        yield
+    finally:
+        for m in reversed(imported_modules):
+            container[Logger].info(f"shutdown module {m.__name__}")
```
kichanyurd, line 228: `LOGGER.info(f"Shutting down module '{m.__name__}'")`

src/parlant/bin/server.py
```diff
@@ -182,6 +186,49 @@ async def create_agent_if_absent(agent_store: AgentStore) -> None:
         await agent_store.create_agent(name=DEFAULT_AGENT_NAME)
 
 
+@asynccontextmanager
+async def load_modules(
+    container: Container,
+    modules: list[str],
+) -> AsyncIterator[None]:
+    if not modules:
```
kichanyurd, line 194: It's bad design to specifically provide a list of modules to this function, then go and do something else inside like read a file as a fallback.

This function should just get the exact modules that need to be run. And this function should **only** be run if the list of modules is non-empty.

Understanding which modules should be run based on config or params should be done in a separate function.

PR #187, opened 2024-12-15
Feature/mc 433 add locks to session store and json database findfind one

mc-dorzo:


src/parlant/core/async_utils.py
```diff
@@ -127,3 +130,28 @@ async def safe_gather(  # type: ignore[misc]
         for future in futures:
             future.cancel()
         raise
+
+
+class RWLock:
```
kichanyurd, line 135: Rename to `ReaderWriteLock`

src/parlant/core/async_utils.py
```diff
@@ -13,10 +13,13 @@
 # limitations under the License.
 
 from __future__ import annotations
-from typing import Any, Coroutine, Iterable, TypeVar, overload
+from contextlib import asynccontextmanager
+from typing import Any, AsyncIterator, Coroutine, Iterable, TypeVar, overload, AsyncContextManager
 import asyncio
 import math
 
+import aiorwlock
```
kichanyurd, line 21: No need for extra space :)

src/parlant/core/sessions.py
```diff
@@ -388,6 +388,8 @@ def __init__(self, database: DocumentDatabase):
         self._event_collection: DocumentCollection[_EventDocument]
         self._inspection_collection: DocumentCollection[_InspectionDocument]
 
+        self._lock = asyncio.Lock()
```
kichanyurd, line 391: Why not use rw lock here?

src/parlant/core/sessions.py
```diff
@@ -677,24 +683,27 @@ async def create_event(
         if not await self._session_collection.find_one(filters={"id": {"$eq": session_id}}):
             raise ItemNotFoundError(item_id=UniqueId(session_id), message="Session not found")
 
-        session_events = await self.list_events(
-            session_id
-        )  # FIXME: we need a more efficient way to do this
-        creation_utc = creation_utc or datetime.now(timezone.utc)
-        offset = len(list(session_events))
+        async with self._lock.writer_lock:
```
kichanyurd, line 686: Why is this a writer lock? And why aren't you locking over `find_one()` as well?

kichanyurd: Oh wait I understand why it's a writer lock. The UI here gave me the impression it's under `read_session()`. But I think you need to include `find_one()` in the locked scope, otherwise you can assume that it exists and then go on to find nothing.

PR #189, opened 2024-12-18
feature/mc-449-rewrite-pytest-stochastics

MCNatsu:
feat(test): new stochastics.json file
feat(test): create basic separation to stable/unstable/experimental
fix(just): updated justfile
fix(lock): sort deps and relock
fix(.gitignore): add entries
feat(ci): Trim and update to new specs


MCNatsu:
oops

PR #196, opened 2024-12-20
Add few shot feature and implement it on guideline_proposer

mc-dorzo:


src/parlant/core/shots.py
```diff
@@ -0,0 +1,45 @@
+from abc import abstractmethod
+from dataclasses import dataclass
+from typing import TypeVar, Generic, Sequence
+
+
+@dataclass
+class Shot:
+    description: str
+    """An explanation of what makes this shot interesting."""
```
kichanyurd, line 9: No period at the end of the description

src/parlant/core/shots.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @abstractmethod
+    def format(self) -> str: ...
```
kichanyurd, line 12: Not good to have format under the shot itself, we need to separate logic from presentation.

src/parlant/core/shots.py
```diff
@@ -0,0 +1,45 @@
...
+        self._shots.insert(index, shot)
+
+    async def list(self) -> Sequence[TShot]:
+        return self._shots
+
+    async def remove(
+        self,
+        shot: TShot,
+    ) -> None:
+        self._shots = [s for s in self._shots if s != shot]
```
kichanyurd, line 42: Why not `_shots.remove(shot)`?

tests/core/engines/alpha/test_guideline_proposer.py
```diff
@@ -223,7 +225,11 @@ def propose_guidelines(
     terms: Sequence[Term] = [],
     staged_events: Sequence[EmittedEvent] = [],
 ) -> Sequence[GuidelineProposition]:
-    guideline_proposer = GuidelineProposer(context.logger, context.schematic_generator)
+    guideline_proposer = GuidelineProposer(
+        context.logger,
+        context.schematic_generator,
+        context.container[ShotCollection[GuidelinePropositionShot]],
```
kichanyurd, line 231: You don't need to provide it to the guideline proposer like this -- it's already accessible to it as part of its module.

tests/core/test_shots.py
```diff
@@ -0,0 +1,76 @@
...
+    container: Container,
+    agent: Agent,
+    customer: Customer,
+) -> None:
+    guideline_proposer = container[GuidelineProposer]
+    shot_collection = container[ShotCollection[GuidelinePropositionShot]]
+
+    guideline_content = GuidelineContent(
+        condition="A user compliments the product",
+        action="Tell him to stop being a bitch",
```
kichanyurd, line 27: Ehh maybe dial it down a bit :P

tests/core/test_shots.py
```diff
@@ -0,0 +1,76 @@
...
+                    applies_score=8,
+                )
+            ]
+        ),
+    )
+
+    await shot_collection.append(guideline_proposition_shot)
+
+    shots = await shot_collection.list()
+    prompt = guideline_proposer._format_prompt(
```
kichanyurd, line 63: We shouldn't access underscore-prefixed methods from outside the class: they're considered private.
If you want to access it, you need to make it public—remove the underscore prefix.

HOWEVER -- I wouldn't go through hoops here to format the prompt. It's delving too deep into implementation details, and the assertions you have there are quite fragile anyway.

Instead of using `_format_prompt`, just expose a `guideline_proposer.shots` property which internally simply returns the module's shots (which is the one you're modifying via the container). Then assert that the added shots are contained in it.

src/parlant/core/engines/alpha/prompt_builder.py
```diff
@@ -238,3 +240,24 @@ def add_staged_events(
             )
 
         return self
+
+    def add_few_shots(
```
kichanyurd, line 244: I don't think we need this. Let @MCBarKar format things however we wishes for each component, manually.
This design is also not great since it ties presentation logic into the shot itself via the `format` method.

src/parlant/core/engines/alpha/guideline_proposer.py
```diff
@@ -55,6 +58,60 @@ class GuidelinePropositionsSchema(DefaultBaseModel):
...
 
 
+@dataclass
+class GuidelinePropositionShot(Shot):
+    interaction_events: Sequence[Event]
+    guidelines: Sequence[GuidelineContent]
+    expected_result: GuidelinePropositionsSchema
+
+    @override
+    def format(self) -> str:
```
kichanyurd, line 68: Make `_format_shot` a member method of `GuidelineProposer`.

src/parlant/core/engines/alpha/prompt_builder.py
```diff
@@ -238,3 +240,24 @@ def add_staged_events(
...
+        self,
+        shots: Sequence[Shot],
+    ) -> PromptBuilder:
+        if shots:
+            self.add_section(
+                name=BuiltInSection.FEW_SHOTS,
+                content="".join(
+                    f"""
+Example #{i}:
```
kichanyurd, line 253: Aren't you missing the separators? Those can be important (e.g. ------------ or ###)

tests/core/test_shots.py
mc-dorzo: haha I wasn't sure what you'll think about it

tests/core/test_shots.py
mc-dorzo: Since the module's shots is actually the ShotCollection, and we have an async function for list the shots, I made the function of `async def shots(self) -> Sequence[GuidelinePropositionShot]` instead of property (unless you want other option i.e. propetry of shots return the ShotCollection / async_property .. )

kichanyurd (approved):

PR #198, opened 2024-12-22
Feature/mc 462 trigger message generation manually

mc-dorzo:


examples/modules/tech_store.py
```diff
@@ -14,12 +15,35 @@ def list_categories(context: ToolContext) -> ToolResult:
...
+
+    async with httpx.AsyncClient(follow_redirects=True, timeout=30) as client:
+        (
+            await client.post(
+                f"http://localhost:8800/sessions/{context.session_id}/events",
+                json={
+                    "kind": "message",
+                    "source": "ai_agent",
+                    "actions": [
+                        {"action": "Tell the user that I'm thinking", "reason": "buy_time"}
```
kichanyurd, line 32: Who is who here? I suggest "Tell the user that you're thinking and will be right back with an answer"

src/parlant/api/sessions.py
```diff
@@ -210,13 +211,33 @@ class SessionCreationParamsDTO(
     ),
 ]
 
+SessionEventCreationParamsActionField: TypeAlias = Annotated[
+    str,
+    Field(
+        description="A single action from the AI agent, such as 'the the user Thinking..' or 'inform the user that more information is coming'",
```
kichanyurd, line 217: Instead of "the the user is Thinking"
`description='A single action that explains what to say; i.e. "Tell the customer that you're thinking and will be right back with an answer."'`

src/parlant/core/application.py
```diff
@@ -141,6 +141,34 @@ async def _process_session(self, session: Session) -> None:
...
+        actions: Sequence[UtteranceRequest],
+    ) -> str:
+        with self._correlator.correlation_scope(generate_id()):
+            await self._background_task_service.restart(
+                self._do_utter_session(session, actions),
+                tag=f"utter-session({session.id})",
+            )
+            return self._correlator.correlation_id
+
+    async def _do_utter_session(
```
kichanyurd, line 156: `_do_utter_session`? :) "Session!"
Maybe rename to just `_do_utter`

src/parlant/core/application.py
```diff
@@ -141,6 +141,34 @@ async def _process_session(self, session: Session) -> None:
             event_emitter=event_emitter,
         )
 
+    async def dispatch_utter_task(
+        self,
+        session: Session,
+        actions: Sequence[UtteranceRequest],
+    ) -> str:
+        with self._correlator.correlation_scope(generate_id()):
+            await self._background_task_service.restart(
```
kichanyurd, line 150: Why is this a background task?
Aren't we awaiting it in the API anyway?

src/parlant/api/sessions.py
```diff
@@ -1307,7 +1337,12 @@ async def _add_agent_message(
 
         session = await session_store.read_session(session_id)
 
-        correlation_id = await application.dispatch_processing_task(session)
+        if params.actions:
+            actions = [utterance_request_dto_to_utterance_request(a) for a in params.actions]
+            correlation_id = await application.dispatch_utter_task(session, actions)
```
kichanyurd, line 1342: Didn't we discuss that the API should return only when the uttering is completed? So that the tool would not be in a rate with the uttering.

If so then why are we dispatching a background task and returning immediately?

src/parlant/core/application.py
kichanyurd: Or remove this entirely if we don't need a bg task (which I think we don't)

src/parlant/core/engines/alpha/engine.py
```diff
@@ -139,6 +139,48 @@ async def process(
...
 
+    @override
+    async def utter(
+        self,
+        context: Context,
+        event_emitter: EventEmitter,
+        actions: Sequence[UtteranceRequest],
+    ) -> bool:
+        interaction_state = await self._load_interaction_state(context)
+        session_id = context.session_id
```
kichanyurd, line 150: Probably no need for this variable

src/parlant/core/engines/alpha/engine.py
```diff
@@ -139,6 +139,48 @@ async def process(
...
+
+        try:
+            with self._logger.operation(f"Uttering actions '{actions}' for session {session_id}"):
```
kichanyurd, line 153: Are we sure this log is correctly displaying the content of the utterance requests?

src/parlant/core/engines/alpha/engine.py
```diff
@@ -139,6 +139,48 @@ async def process(
             self._logger.critical(f"Critical processing error: {traceback.format_exception(exc)}")
             raise
 
+    @override
+    async def utter(
+        self,
+        context: Context,
+        event_emitter: EventEmitter,
+        actions: Sequence[UtteranceRequest],
```
kichanyurd, line 147: `requests: Sequence[UtteranceRequests]`

src/parlant/core/engines/alpha/engine.py
```diff
@@ -139,6 +139,48 @@ async def process(
...
+        interaction_state = await self._load_interaction_state(context)
+        session_id = context.session_id
+
+        try:
+            with self._logger.operation(f"Uttering actions '{actions}' for session {session_id}"):
+                await self._do_utter(context, interaction_state, event_emitter, actions)
+            return True
+
+        except asyncio.CancelledError:
+            self._logger.warning(f"Uttering actions for session {session_id} was cancelled.")
```
kichanyurd, line 158: "Uttering for session... was cancelled" not "Uttering actions...was cancelled"

src/parlant/core/engines/alpha/engine.py
```diff
@@ -139,6 +139,48 @@ async def process(
...
+            return False
+
+        except Exception as exc:
+            formatted_exception = traceback.format_exception(type(exc), exc, exc.__traceback__)
+            self._logger.error(
+                f"Error during uttering actions for session {session_id}: {formatted_exception}"
```
kichanyurd, line 164: "Error during uttering for session..." not "Error during uttering actions for session..."

src/parlant/core/engines/alpha/engine.py
```diff
@@ -139,6 +139,48 @@ async def process(
...
+                    "status": "error",
+                    "acknowledged_offset": interaction_state.last_known_event_offset,
+                    "data": {"exception": formatted_exception},
+                },
+            )
+            return False
+
+        except BaseException as exc:
+            self._logger.critical(
+                f"Critical error during uttering actions for session {session_id}: "
```
kichanyurd, line 179: Same here

src/parlant/core/engines/alpha/engine.py
```diff
@@ -396,6 +438,106 @@ async def _do_process(
...
+
+            ordinary_guideline_propositions = await self._utter_requests_to_guideline_propositions(
+                actions
+            )
+
+            await event_emitter.emit_status_event(
+                correlation_id=self._correlator.correlation_id,
+                data={
+                    "acknowledged_offset": interaction.last_known_event_offset,
+                    "status": "processing",
```
kichanyurd, line 481: Status type here should be "typing" not "processing", since we are literally just typing a message and not running any tools.

src/parlant/core/engines/alpha/engine.py
```diff
@@ -623,6 +765,36 @@ async def _load_relevant_terms(
             )
         return []
 
+    async def _utter_requests_to_guideline_propositions(
+        self,
+        actions: Sequence[UtteranceRequest],
+    ) -> Sequence[GuidelineProposition]:
+        def utterance_to_proposition(utterance: UtteranceRequest) -> GuidelineProposition:
+            ratioanles = {
```
kichanyurd, line 773: rationales

src/parlant/core/engines/alpha/engine.py
```diff
@@ -623,6 +765,36 @@ async def _load_relevant_terms(
...
+            }
+
+            conditions = {
+                UtteranceReason.BUY_TIME: "always",
+                UtteranceReason.FOLLOW_UP: "always",
+            }
+
+            return GuidelineProposition(
+                guideline=Guideline(
+                    id=GuidelineId(""),
```
kichanyurd, line 785: Maybe include at least some indicator here so that it doesn't look like a bug: `GuidelineId("<utterance-request-{i}>")` where `i` is received by enumerating `actions` (which BTW should be named `requests`)

src/parlant/core/engines/alpha/engine.py
```diff
@@ -623,6 +765,36 @@ async def _load_relevant_terms(
             )
         return []
 
+    async def _utter_requests_to_guideline_propositions(
+        self,
+        actions: Sequence[UtteranceRequest],
```
kichanyurd, line 770: `requests: ...`

tests/core/engines/alpha/steps/engines.py
```diff
@@ -140,3 +140,30 @@ def when_messages_are_emitted(
     assert all(e is not None for e in message_event_generator_results[0].events)
 
     return list(cast(list[EmittedEvent], message_event_generator_results[0].events))
+
+
+@step(when, "uttring actions", target_fixture="emitted_events")
+def when_uttring_actions(
```
kichanyurd, line 146: uttering

kichanyurd: No need "uttring actions"
@step(when, "uttering is triggered", ...
def when_uttering_is_triggered

tests/core/engines/alpha/steps/events.py
```diff
@@ -193,6 +194,22 @@ def given_a_flagged_customer_message(
     return session.id
 
 
+@step(
+    given,
+    parsers.parse('an utterance with action of "{action}" and reason of "{reason}"'),
+)
+def given_an_utterance(
```
kichanyurd, line 201:
```@step(..., 'an utterance request "{action}", to {reason}')
def given_an_utterance_request```

where in the BDD "reason" shouldn't be in quotes, and should be either "follow up with the customer" or "buy time"

Here's what usage would look like:
`Given an utterance request "Ask if they got your message", to follow up with the customer`
`Given an utterance request "Say you'll be late", to buy time`

PR #226, opened 2025-01-07
Feature/mc 514 consume logs from client side

mc-dorzo:
CLI Logging Commands:
    - Added a log command group to the CLI, allowing users to view logs in real-time with optional filters.
    - Support for filtering logs by specific components:
    - --guideline-proposer (-g)
    - --tool-caller (-t)
    - --message-event-generator (-m)
    - Support for logical operators (AND/OR) between multiple filters.
    - Added support for additional pattern matching on log messages.
    - Example CLI commands:
```
# Streaming all logs
parlant log

# Filter logs by component
parlant log --guideline-proposer

# Combine filters using AND
parlant log -g Evaluation --operator AND

# Add a pattern filter
parlant log --tool-caller [ToolResult]
```


src/parlant/bin/client.py
```diff
@@ -846,6 +847,39 @@ def delete_tag(ctx: click.Context, tag_id: str) -> None:
...
+            sub_socket = context.socket(zmq.SUB)
+            sub_socket.connect(f"{ctx.obj.log_server_address}")
+
+            sub_socket.setsockopt_string(zmq.SUBSCRIBE, "")
+
+            rich.print(Text("Streaming logs...", style="bold yellow"))
+
+            while True:
+                try:
+                    message = cast(dict[str, Any], sub_socket.recv_json(flags=zmq.NOBLOCK))
```
kichanyurd, line 863: Why are we using `NOBLOCK` here?
And then sleeping for 0.01s?
That's a lot of idle spinning

src/parlant/bin/client.py
```diff
@@ -1984,10 +2033,21 @@ class Config:
         metavar="ADDRESS[:PORT]",
         default="http://localhost:8800",
     )
+    @click.option(
+        "--log-server-address",
```
kichanyurd, line 2037: How it's written now we would have to write the following to access a remote server:

`parlant --server http://1.2.3.4:8800 --log-server-address tcp://1.2.3.4:8799 [...]`

Instead let's assume that the log server is on the same host as the server itself, and just allow specifying a different port. So the option would be `--log-port` and we'd use the host name from the `--server` option.

src/parlant/core/engines/alpha/guideline_proposer.py
```diff
@@ -233,7 +233,7 @@ async def _process_guideline_batch(
             guideline = guidelines_dict[int(proposition.guideline_number)]
```
kichanyurd, line 233: Let's add 2 logs here:

`[GuidelineProposer][Prompt]` (debug)
`[GuidelineProposer][PromptOutput]` (debug) --> output all of the response's content as indented JSON

src/parlant/core/engines/alpha/message_event_generator.py
```diff
@@ -180,11 +180,13 @@ async def generate_events(
...
                     else:
-                        self._logger.debug("Skipping response; no response deemed necessary")
+                        self._logger.debug(
+                            "[MessageEventGenerator] Skipping response; no response deemed necessary"
+                        )
                         return [MessageEventGenerationResult(generation_info, [])]
                 except Exception as exc:
                     self._logger.warning(
-                        f"Generation attempt {generation_attempt} failed: {traceback.format_exception(exc)}"
+                        f"[MessageEventGenerator] Generation attempt {generation_attempt} failed: {traceback.format_exception(exc)}"
```
kichanyurd, line 189: In MEG let's also add the log
`[MessageEventGenerator][Completion]` --> logging all of the generated content as indented JSON.

And BTW let's do the same for ToolCaller too... so we have `[Prompt]` and `[Completion]` for all components.

src/parlant/core/engines/alpha/message_event_generator.py
```diff
@@ -171,7 +171,7 @@ async def generate_events(
                     )
 
                     if response_message is not None:
-                        self._logger.debug(f'Message production result: "{response_message}"')
+                        self._logger.debug(f'[MessageEventGenerator][Result]: "{response_message}"')
```
kichanyurd, line 174: Rename, this isn't just "result", this is `[MessageEventGenerator][GeneratedMessage]`

src/parlant/core/engines/alpha/message_event_generator.py
```diff
@@ -518,7 +522,7 @@ async def _generate_response_message(
             and not final_revision.is_repeat_message
         ):
             self._logger.warning(
-                f"PROBLEMATIC MESSAGE EVENT PRODUCER RESPONSE: {final_revision.content}"
+                f"[MessageEventGenerator] PROBLEMATIC MESSAGE EVENT PRODUCER RESPONSE: {final_revision.content}"
```
kichanyurd, line 525: I think you'll have to rebase here as this part was changed already

src/parlant/core/engines/alpha/tool_caller.py
```diff
@@ -555,7 +557,7 @@ async def _run_inference(
 
         log_message = json.dumps(inference.content.model_dump(mode="json"), indent=2)
 
-        self._logger.debug(f"Tool call request results: {log_message}")
+        self._logger.debug(f"[ToolCaller][RequestResults]: {log_message}")
```
kichanyurd, line 560: Rename to `[ToolCaller][Completion]`

src/parlant/core/engines/alpha/message_event_generator.py
```diff
@@ -142,7 +142,7 @@ async def generate_events(
                 shots=await self.shots(),
             )
 
-            self._logger.debug(f"Message production prompt:\n{prompt}")
+            self._logger.debug(f"[MessageEventGenerator][Prompt]:\n{prompt}")
```
kichanyurd, line 145: I see you sometimes have a colon (:) after the `[..]` part and sometimes don't.
Let's be consistent: go over all instances and let's go with **no colon**. Just a space after the last bracket (in this case a newline is fine instead of the space since it's a multiline prompt)

src/parlant/core/logging.py
```diff
@@ -179,3 +181,119 @@ def __init__(
...
+    @contextmanager
+    def _nested_contexts(contexts: Sequence[AbstractContextManager[Any]]) -> Iterator[None]:
+        exits = []
+        try:
+            for context in contexts:
+                exits.append(context.__enter__())
+            yield
+        finally:
+            for context in reversed(contexts):
+                context.__exit__(None, None, None)
```
kichanyurd, line 299: There's a bug here. What if we got an exception? We'll be calling `__exit__` here without one.

Why not use `ExitStack` do write this cleanly?

tests/core/stable/test_loggers.py
```diff
@@ -0,0 +1,60 @@
...
+    return messages
+
+
+async def test_that_zmq_logger_emits_logs(zmq_logger: ZMQLogger) -> None:
+    context = zmq.asyncio.Context.instance()
+    sub_socket = context.socket(zmq.SUB)
+    sub_socket.connect(f"tcp://localhost:{PARLANT_LOG_PORT}")
+    sub_socket.setsockopt_string(zmq.SUBSCRIBE, "")
+
+    await asyncio.sleep(1)
```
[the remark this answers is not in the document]
mc-dorzo: It need to to connect. otherwise we the messages are sent without the subscriber receiving in

src/parlant/bin/client.py
mc-dorzo: removed the block. I thought in the begining it will be async and maybe pulling multiple topics. nvm. ^^

PR #232, opened 2025-01-15
Introduced DCO related documentation and configuration

MCNatsu:
This PR contains the required docs and configs required for initiating the DCO workflow with [this dco bot](https://github.com/apps/dco)

## NOTICE:
* [ ] TODO: install the bot on the repo
* [ ] NOTE: remediation will only be accepted after the `dco.yml` file is on the `main` branch.


MCNatsu:
There's a chance that when you enable the both, it will whine about all our previously unsigned commits. If this happens, there's a remediation msg you can copy paste from its report (I'll be on it).

PR #233, opened 2025-01-15
Allow only specifying guideline condition or action when updating gui…

mc-dorzo:
…deline from CLI

### DIFF
--- a/src/parlant/bin/client.py
+++ b/src/parlant/bin/client.py
@@ -2476,7 +2476,9 @@ async def async_main() -> None:
         check: bool,
         connect: bool,
     ) -> None:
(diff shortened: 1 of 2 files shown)


src/parlant/bin/client.py
```diff
@@ -2468,6 +2476,8 @@ def guideline_update(
         check: bool,
         connect: bool,
     ) -> None:
+        assert condition or action
```
kichanyurd, line 2479: This isn't a good way to do it. If someone runs it with neither `--action` or `--condition`, instead of getting a clear error message that at least one is required, they will get an assertion error.

PR #250, opened 2025-01-22
Feature/mc 534 enhancement log inspection for developer chat messages

mc-dorzo:


src/parlant/core/engines/alpha/guideline_proposer.py
```diff
@@ -251,29 +232,22 @@ async def _process_guideline_batch(
...
-                f"  ShouldReapply: {proposition.guideline_should_reapply}\n"
-                f'  ReapplicationRationale: "{proposition.guideline_previously_applied_rationale}"'
-            )
-
             if (proposition.applies_score >= 6) and (
                 (proposition.guideline_previously_applied == "no")
                 or proposition.guideline_should_reapply
             ):
+                self._logger.debug(
+                    f"[GuidelineProposer][Completion][Activated] {proposition.model_dump_json(indent=2)}"
```
kichanyurd, line 248: Maybe add a newline *instead* of the space after [Activated]?
Same for all the other messages in this file and in other files.

PR #259, opened 2025-01-28
Feature/mc 535 enhancement migrate store versions on startup

mc-dorzo:
introduces a migration mechanism for document stores and a --migrate CLI flag to streamline the management of store version upgrades during server startup. Together, these enhancements ensure compatibility, reduce manual intervention, and provide clear pathways for handling migrations.

Migration Logic Overview:

1. New Store (No Metadata Exists):
	•	No Migration Required: As this is a new store, there are no existing documents to migrate.
	•	Action: A metadata document is created

2. Existing Store:
    2.1: No Metadata Exists:
        2.1.1: Documents Are in the Same Version as the Store:
          No Migration Required: Documents are already aligned with the store version.
	  Action: A metadata document is created to reflect the store version.
        2.1.2: Documents Are Not in the Same Version as the Store:
  	  Migration Required: Mismatch between the document version and the store version.
	  Action: A metadata document is created with the document version, and migration is triggered.
    2.2: Metadata Already Exists:
	2.2.1: Metadata Version Matches the Store Version:
	  No Migration Required: Metadata and documents are aligned with the store version.
	2.2.2: Metadata Version Does Not Match the Store Version:
	  Migration Required: The metadata version indicates a discrepancy with the store version.

CLI Flag: --migrate

This PR also introduces the --migrate CLI flag to streamline migration processes during server startup.

Key Behaviors:
	•	If the --migrate flag is enabled:
	•	The server automatically executes the required migration process, upgrading the store and documents to the latest version.
	•	This ensures compatibility without requiring additional user input.
	•	If the --migrate flag is not specified:
	•	If a migration is required, the server raises a MigrationError.
	•	Critical instructions are logged, and the server exits gracefully to prevent unintentional execution with incompatible data.


src/parlant/adapters/vector_db/chroma.py
```diff
@@ -87,6 +62,65 @@ async def __aexit__(
     ) -> None:
         pass
 
+    async def load_documents_to_chroma_collection_with_loader(
```
kichanyurd, line 65: If not public, should be prefixed with _

src/parlant/core/agents.py
```diff
@@ -95,23 +99,67 @@ class _AgentDocument(TypedDict, total=False):
...
     VERSION = Version.from_string("0.1.0")
 
-    def __init__(
-        self,
-        database: DocumentDatabase,
-    ):
+    def __init__(self, database: DocumentDatabase, migrate: bool = True):
         self._database = database
         self._collection: DocumentCollection[_AgentDocument]
+        self._meta_collection: DocumentCollection[_MetadataDocument]
```
kichanyurd, line 113: In all stores, rename to `_metadata_collection`. Reason: "Meta data" means "data about data". Meaning of "meta collection" means "collection about collections". These are separate meanings. "Meta" isn't simply a shorthand for "metadata".

src/parlant/core/agents.py
```diff
@@ -95,23 +99,67 @@ class _AgentDocument(TypedDict, total=False):
...
 
+        self._migrate = migrate
         self._lock = ReaderWriterLock()
 
+    async def _meta_document_loader(self, doc: BaseDocument) -> Optional[_MetadataDocument]:
+        if doc["version"] == "0.1.0":
+            return cast(_MetadataDocument, doc)
+
+        if not self._migrate:
+            raise VersionMismatchError(
```
kichanyurd, line 123: What exactly is the difference between `VersionMismatchError` and `MigrationRequiredError`?

Wouldn't migration always be required if there is a version mismatch?
Wouldn't there always be a version mismatch if a migration was required?

Why are there these two different errors that seem to point to the exact same thing?

src/parlant/core/persistence/document_database.py
```diff
@@ -55,6 +57,10 @@ class DeleteResult(Generic[TDocument]):
     deleted_document: Optional[TDocument]
 
 
+async def noop_loader(doc: BaseDocument) -> BaseDocument:
```
kichanyurd, line 60: No-Op usually means "do nothing" but in this case we are doing something: returning the argument as-is.

This is commonly known as an "identity function". For this reason, I suggest calling this `identity_loader`.

kichanyurd: https://en.wikipedia.org/wiki/Identity_function

mc-dorzo: I knew that from math ^^ awesome observation

src/parlant/core/agents.py
mc-dorzo: made one of MigrationError, since we now have only the case of version mismatch (before that I decided to force the user to migrate in case version match but no metadata.)

PR #279, opened 2025-02-15
Feature/mc 552 enhancement clearer rate limit induced error reporting

mc-dorzo:


src/parlant/adapters/nlp/openai.py
```diff
@@ -119,12 +132,15 @@ async def _do_generate(
...
-            )
+            try:
+                response = await self._client.beta.chat.completions.parse(
+                    messages=[{"role": "developer", "content": prompt}],
+                    model=self.model_name,
+                    response_format=self.schema,
+                    **openai_api_arguments,
+                )
+            except RateLimitError as e:
+                raise RateLimitError(RATE_LIMIT_ERROR_MESSAGE, response=e.response, body=e.body)
```
kichanyurd, line 143: Not sure if it makes sense to raise a new exception with this message. Maybe just log.error() with the message and re-raise the same exception

PR #283, opened 2025-02-17
Feature/mc 532 enhancement allow disabling guidelines

mc-dorzo:


src/parlant/core/guidelines.py
```diff
@@ -202,13 +233,17 @@ async def create_guideline(
     async def list_guidelines(
         self,
         guideline_set: str,
+        show_disabled: bool = False,
```
kichanyurd, line 236: This isn't the place to do it.
It's not like with events where they're *deleted* so we stop listing them.
Here the guideline is disabled but it should still be listed everywhere like API, CLI, and it shouldn't take extra effort or flags to do so.

It's only the engine itself which must filter out disabled guidelines.

src/parlant/bin/client.py
```diff
@@ -2672,17 +2729,26 @@ def guideline_view(
         metavar="ID",
         required=False,
     )
+    @click.option(
+        "--show-disabled",
```
kichanyurd, line 2733: Let's change this to `--hide-disabled` and set it to False by default.

src/parlant/core/guidelines.py
```diff
@@ -107,18 +109,42 @@ class _GuidelineDocument(TypedDict, total=False):
...
-        self._collection: DocumentCollection[_GuidelineDocument]
+        self._collection: DocumentCollection[_GuidelineDocumentV0_2_0]
         self._allow_migration = allow_migration
         self._lock = ReaderWriterLock()
 
-    async def _document_loader(self, doc: BaseDocument) -> Optional[_GuidelineDocument]:
+    async def _document_loader(self, doc: BaseDocument) -> Optional[_GuidelineDocumentV0_2_0]:
         if doc["version"] == "0.1.0":
-            return cast(_GuidelineDocument, doc)
+            doc_v0_1_0 = cast(_GuidelineDocument, doc)
```
kichanyurd, line 133: The naming should be different.

`_GuidelineDocument` should always point to the latest version.
It's the old `_GuidelineDocument` that should be renamed to `_GuidelineDocument_V0_1_0`.

PR #313, opened 2025-03-09
feat(adapters): add `MongoDocumentDatabase`

Relvox:
This PR Introduces a `MongoDocumentDatabase` adapter (in `adapters/db/mongo_db.py`), based on the discussion in https://github.com/emcie-co/parlant/issues/307 .

### Notable implementation details
* creating a `MongoDocumentDatabase` requires providing a live `AsyncMongoClient` instance.
* `get_or_create_collection` is just a wrapper for `get_collection` since get also handles creation if it does not exist.
* following the `failed_migrations` protocol for when the `document_loader` fails on a document.
* `update_one` and `delete_one` of `MongoDocumentCollection` need to call an extra `find_one` in order to have a value to return in their respective result objects.
* Added tests to `tests/adapters/db/test_mongodb.py` based on corresponding `json_file` tests.

### N.B.

* current version on develop doesn't pass mypy `pre-push` ... so I had to `--no-verify` it, after making sure none of my files were the offending ones.
* also note that without a valid `OPENAI_API_KEY` there's hardly any points in running the test suite, it all fails.

### Questions

* I guess we can make `create_collection` also call on `get_collection` with an `identity_loader` to make it even smaller? Or should get_or_create not call `self` and explicitly match the called action (even though mongo doesn't care)?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,6 +62,7 @@ google-genai = { version = "^1.2.0", optional = true }
 together = { version = "^1.2.12", optional = true }
 torch = { version = "^2.5.1", optional = true }
 transformers = { version = "^4.46.2", optional = true }
+pymongo = "^4.11.1"
 
 [tool.poetry.group.dev.dependencies]
 ipython = "^8.26.0"
--- a/src/parlant/adapters/db/mongo_db.py
+++ b/src/parlant/adapters/db/mongo_db.py
@@ -80,26 +80,36 @@ class MongoDocumentDatabase(DocumentDatabase):
         )
 
         flag_any_failed = False
+        failed_migrations_collection_name = f"{self.database_name}_{name}_failed_migrations"
         collection_existing_documents = result_collection.find({})
-        failed_migration_collection = await self.create_collection("failed_migrations", schema)
+        if failed_migrations_collection_name in await self._database.list_collection_names():
+            self._logger.info(f"deleting old `{failed_migrations_collection_name}` collection")
+            await self.delete_collection(failed_migrations_collection_name)
+
+        failed_migration_collection = await self.create_collection(
+            failed_migrations_collection_name, schema
+        )
         for doc in await collection_existing_documents.to_list():
             try:
                 if loaded_doc := await document_loader(doc):
                     await result_collection.replace_one(doc, loaded_doc)
                 else:
                     flag_any_failed = True
-                    self._logger.warning(f'Failed to load document "{doc}"')
+                    self._logger.warning(f'failed to load document "{doc}"')
                     await failed_migration_collection.insert_one(doc)
                     await result_collection.delete_one(doc)
             except Exception as e:
                 flag_any_failed = True
                 self._logger.error(
-                    f"Failed to load document '{doc}' with error: {e}. Added to failed migrations collection."
+                    f"failed to load document '{doc}' with error: {e}. Added to `{failed_migrations_collection_name}` collection."
                 )
                 await failed_migration_collection.insert_one(doc)
 
         if not flag_any_failed:
-            await self.delete_collection("failed_migrations")
+            self._logger.info(
+                f"deleting `{failed_migrations_collection_name}` collection as no migrations failed this run"
+            )
+            await self.delete_collection(failed_migrations_collection_name)
 
         self._collections[name] = MongoDocumentCollection(self, result_collection)
         return self._collections[name]
--- a/src/parlant/adapters/db/mongo_db.py
+++ b/src/parlant/adapters/db/mongo_db.py
@@ -1,0 +1,175 @@
+# Copyright 2025 Emcie Co Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


src/parlant/adapters/db/mongo_db.py
```diff
@@ -0,0 +1,184 @@
...
+)
+from pymongo import AsyncMongoClient
+from pymongo.asynchronous.database import AsyncDatabase
+from pymongo.asynchronous.collection import AsyncCollection
+
+
+class MongoDocumentDatabase(DocumentDatabase):
+    def __init__(
+        self,
+        mongo_client_async: AsyncMongoClient[Any],
```
kichanyurd, line 37: Can rename param to just `mongo_client` (async implied)

src/parlant/adapters/db/mongo_db.py
```diff
@@ -0,0 +1,184 @@
...
+                    self._logger.warning(f'Failed to load document "{doc}"')
+                    failed_migrations.append(doc)
+                    await result_collection.delete_one(doc)
+            except Exception as e:
+                self._logger.error(
+                    f"Failed to load document '{doc}' with error: {e}. Added to failed migrations collection."
+                )
+                failed_migrations.append(doc)
+
+        if len(failed_migrations):
```
kichanyurd, line 98: It could be that the process dies (or is interrupted) before reaching this stage, but after running all the `delete_one()` calls. Therefore, I'd suggest explicitly creating the `failed_migrations` collection ahead of time and immediately insert to it ***right before*** deleting the document from the actual collection. This is the safest route.

Then, if it turns out we didn't have *any* failed migrations, we can finally delete the speculatively-created `failed_migrations` collection.

src/parlant/adapters/db/mongo_db.py
```diff
@@ -0,0 +1,184 @@
...
+        result = await mongo_cursor.to_list()
+        await mongo_cursor.close()
+        return result
+
+    async def find_one(self, filters: Where) -> TDocument | None:
+        result = await self._collection.find_one(filters)
+        return result
+
+    async def insert_one(self, document: TDocument) -> InsertResult:
+        await self._collection.insert_one(document)
```
kichanyurd, line 156: Doesn't pymongo return its own `acknowledged` value? We could relay it here instead of guessing it to be True

pyproject.toml
```diff
@@ -62,6 +62,7 @@ google-genai = { version = "^1.2.0", optional = true }
 together = { version = "^1.2.12", optional = true }
 torch = { version = "^2.5.1", optional = true }
 transformers = { version = "^4.46.2", optional = true }
+pymongo = "^4.11.1"
```
kichanyurd, line 65: I think it's safer in PRs to only change `pyproject.toml` and not `poetry.lock` as solving merge conflicts with the latter is so much of a pain that it's better to just do it manually directly on `develop` after merging. So on that note, could you please remove `poetry.lock` from this PR?

src/parlant/adapters/db/mongo_db.py
Relvox: Yes, thank you :)

src/parlant/adapters/db/mongo_db.py
Relvox: Aye, will fix

Relvox: what should we do if there already is a `failed_migrations` collection in the database?

Relvox: each `failed_migration` collection will now have a prefix of the `database_name_` and `collection_name_` to make ensure that unrelated calls to `get_collection` do no interfere with each others failed migrations.

src/parlant/adapters/db/mongo_db.py
Relvox: also, done

pyproject.toml
Relvox: fact based and done

Relvox:
rebased and added some efficiency ;)

The optimization (only creating the failed migration collection if any migrations fail) is mainly targeted at using mongo atlas. When the mongo server **isn't local** it took `parlant-server` around _60 seconds_ to load due to creating and dropping collections. now it only takes _20 seconds_.

src/parlant/adapters/db/mongo_db.py
kichanyurd: IIUC, sounds good 👍

tests/adapters/db/test_mongodb.py
```diff
@@ -0,0 +1,1117 @@
...
+        ({"name": "Test Agent", "description": "You are a test agent"}),
+    ],
+)
+async def test_agent_creation(
+    context: _TestContext,
+    test_mongo_client: Optional[AsyncMongoClient[Any]],
+    test_database_name: str,
+    agent_configuration: dict[str, Any],
+) -> None:
+    if test_mongo_client is None:
```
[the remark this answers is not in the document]
Relvox: Completely agree, that comment was missing a `TODO:` tag 😅

Relvox:
fixed tests and rebased

PR #318, opened 2025-03-11
feat: Support for Qdrant

Anush008:
## NOTE

I am in the process of working out some remaining test cases.

## Description

This PR adds support for Qdrant - https://qdrant.tech/ to be used as a vector database in Parlant.


kichanyurd:
Awesome! Nice initiative!

kichanyurd:
@Anush008 let us know when it's ready. Leaving this open in the meantime.

anush008:
Ready for review.
The implementation is 400 lines. Rest is test code.

tests/adapters/db/test_qdrant.py
```diff
@@ -0,0 +1,444 @@
...
+@fixture
+async def qdrant_database(context: _TestContext) -> AsyncIterator[QdrantDatabase]:
+    async with create_database(context) as qdrant_database:
+        yield qdrant_database
+
+
+def create_database(context: _TestContext) -> QdrantDatabase:
+    return QdrantDatabase(
+        logger=context.container[Logger],
+        client=QdrantClient(path=context.home_dir),
```
Anush008, line 86: We use the local mode for tests instead of running the Rust based Qdrant server.
https://github.com/qdrant/qdrant-client?tab=readme-ov-file#local-mode

kichanyurd: Hey @Anush008 can you please remove the `poetry.lock` changes from your PR? Otherwise it'll be hard to find the chance to merge without conflicts

anush008:
Hey @kichanyurd. Just bumping this PR. Please take a look when possible.

kichanyurd:
@Anush008 Got it! Just so I understand, do you have an immediate use case for Qdrant on Parlant?
Asking because if so we'd also need to add the option to start the server up with Qdrant, in addition to adding the implementation (which you did here).

If it's not urgent, I'd say we could probably look into merging it properly in about 4 weeks.
If it is urgent, then we need a few changes:

- Add support in `bin/server.py` to start the server up with qdrant instead of chroma
- Change `pyproject.toml` to add `qdrant` as an optional extra package (`parlant[qdrant]`) as I don't want to drag in the dependencies when people aren't using qdrant

LMK WYT

anush008:
Hey @kichanyurd.
Not urgent as such. In the meantime, could you confirm whether the migration script at `src/parlant/bin/prepare_migration.py` also needs to be updated to use Qdrant, or is it there solely for backward compatibility?

kichanyurd:
@mc-dorzo  tagging as you're the expert on the migration script

mc-dorzo:
Hey @Anush008!

Currently, the schema migration script — a.k.a. parlant-prepare-migration — is implemented without configuration options like setting the source and target database types. It’s currently scoped to run within the same database.

I believe the intended approach here is for the schema migrations handled by this script to remain consistent within the same DB type. It shouldn’t be responsible for handling database-to-database transfers.

That said, since we want to move forward with this PR, you have two options IMHO:
	1.	The clean option: Create a new dedicated script for export-import — transferring data from one DB type to another. (Note: for now, you only need to handle the glossary DB, since that’s the only one we use for vector storage.)
	2.	Quick patch option: Add a --qdrant flag to both bin/server.py and bin/prepare-migration.py. When this flag is used, it will trigger the migration that moves the glossary DB from Chroma (our default) to Qdrant.

You can refer to the migrate_relationships_0_2_0_to_0_3_0 function as an example — I used it to move data between two JSONFileDocumentDatabase instances. In your case, you can adapt it to move from ChromaDatabase to QdrantDatabase.
(Also, could you please move the QdrantDatabase class above QdrantCollection in the file, to stay consistent with how we’ve ordered things in other implementations?)

Right now, our migrations go from version to version using the @register_migration decorator. For DB-to-DB transfers, you’ll need to implement a register_transfer decorator that handles the source and target DB types, and make sure it’s integrated into detect_required_migrations.

That’s a lot to take in! If anything is unclear, feel free to ask me here or ping me on Discord. We can also hop on Zoom and pair program if that’s easier! (:

Pro-Sifat-Hasan:
@Anush008 This has already been implemented, so feel free to close this PR.

PR #325, opened 2025-03-25
Feature/mc 596 enhancement guideline specific optimizations

mc-dorzo:


src/parlant/core/engines/alpha/engine.py
```diff
@@ -717,13 +718,15 @@ async def _load_matched_guidelines(
 
         # Step 2: Filter the best matches out of those.
         matching_result = await self._guideline_matcher.match_guidelines(
-            agent=context.agent,
-            customer=context.customer,
+            context=GuidelineMatchingContext(
```
kichanyurd, line 721: I think context should be internal to the guideline matcher module. Not visible to the engine.

src/parlant/core/engines/alpha/guideline_matcher.py
```diff
@@ -102,159 +101,47 @@ def matches(self) -> Sequence[GuidelineMatch]:
...
+class GuidelineMatchingBatchResult(DefaultBaseModel):
+    matches: Sequence[GuidelineMatch]
+    generation_info: GenerationInfo
 
-    async def match_guidelines(
+
+class GuidelineMatchingBatch(ABC):
+    @abstractmethod
+    async def process(self) -> GuidelineMatchingBatchResult:
+        pass
```
kichanyurd, line 112: Replace `pass` with `...`

src/parlant/core/engines/alpha/guideline_matcher.py
```diff
@@ -102,159 +101,47 @@ def matches(self) -> Sequence[GuidelineMatch]:
...
-                customer,
-                guidelines,
-                context_variables,
-                interaction_history,
-                terms,
-                staged_events,
-            )
+        context: GuidelineMatchingContext,
+    ) -> Sequence[GuidelineMatchingBatch]:
+        pass
```
kichanyurd, line 122: Replace `pass` with `...` for abstract methods

src/parlant/core/engines/alpha/guideline_matcher.py
```diff
@@ -494,13 +373,127 @@ def _build_prompt(
...
         return builder
 
 
+class GenericGuidelineMatching(GuidelineMatchingStrategy):
+    def __init__(
+        self,
+        logger: Logger,
+        schematic_generator: SchematicGenerator[GenericGuidelineMatchesSchema],
+    ) -> None:
+        self._generic_guideline_matching_batch = partial(
```
kichanyurd, line 389: Kind of weird that it's a partial and not a helper function.

In any case, it's an action so it should be a verb: `self._create_batch(...)` but I think you should just make it a method explicitly

src/parlant/core/engines/alpha/guideline_matcher.py
```diff
@@ -494,13 +373,127 @@ def _build_prompt(
...
+
+class GuidelineMatcher:
+    def __init__(
+        self,
+        logger: Logger,
+        schematic_generator: SchematicGenerator[GenericGuidelineMatchesSchema],
+    ) -> None:
+        self._logger = logger
+        self._schematic_generator = schematic_generator
+        self.strategy_picker = self._default_strategy_picker
```
kichanyurd, line 442: Where is `self._default_strategy_picker` defined?

PR #327, opened 2025-03-27
Added precedence to ToolParameterOption / MissingToolData

mc-kfir:
Added precedence to ToolParameterOption / MissingToolData
This field is used in order to separate completion requests (from the user) of missing parameters into smaller groups

### DIFF
--- a/src/parlant/core/engines/alpha/engine.py
+++ b/src/parlant/core/engines/alpha/engine.py
@@ -36,7 +36,7 @@ from parlant.core.engines.alpha.utterance_selector import UtteranceSelector
 from parlant.core.engines.alpha.message_event_composer import (
     MessageEventComposer,
 )
-from parlant.core.engines.alpha.tool_caller import ToolInsights
+from parlant.core.engines.alpha.tool_caller import ToolInsights, MissingToolData
 from parlant.core.guidelines import Guideline, GuidelineId, GuidelineContent
 from parlant.core.glossary import Term
 from parlant.core.sessions import (
@@ -230,6 +230,13 @@ class AlphaEngine(Engine):
 
             if not await self._hooks.call_on_generating_messages(context):
                 return
+            
+            # Filter missing tool parameters
+            context.state.tool_insights = ToolInsights(
+                missing_data=await self._filter_missing_tool_parameters(
+                    context.state.tool_insights.missing_data
+                )
+            )
 
             # Money time: communicate with the customer given
             # all of the information we have prepared.
@@ -887,6 +894,20 @@ class AlphaEngine(Engine):
             variable=variable,
             key=key,
         )
+
+    async def _filter_missing_tool_parameters(
+        self, missing_parameters: Sequence[MissingToolData]
+    ) -> Sequence[MissingToolData]:
+        # Precedence 0 is the default, so minimal precedence values start from 1
+        positive_precedence = [
+            m.precedence
+            for m in missing_parameters
+            if m.precedence is not None and m.precedence > 0
+        ]
+        if positive_precedence is None:
+            return missing_parameters
+
+        return [m for m in missing_parameters if m.precedence == min(positive_precedence)]
 
 
 # This is module-level and public for isolated testability purposes.
--- a/src/parlant/core/engines/alpha/engine.py
+++ b/src/parlant/core/engines/alpha/engine.py
@@ -230,7 +230,7 @@ class AlphaEngine(Engine):
 
             if not await self._hooks.call_on_generating_messages(context):
                 return
-            
+
             # Filter missing tool parameters
             context.state.tool_insights = ToolInsights(
                 missing_data=await self._filter_missing_tool_parameters(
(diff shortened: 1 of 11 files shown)


tests/core/common/engines/alpha/steps/events.py
```diff
@@ -527,3 +527,19 @@ def then_the_tool_calls_event_contains_call(
...
+    emitted_events: list[EmittedEvent],
+    number_of_missing: int,
+) -> None:
+    message_event = next(e for e in emitted_events if e.kind == "message")
+    message = cast(MessageEventData, message_event.data)["message"]
+
+    assert context.sync_await(
+        nlp_test(
+            context=f"Here's a message from an AI agent to a customer, in the context of a conversation: {message}",
+            condition=f"The message mentions less than {number_of_missing} missing parameters",
```
kichanyurd, line 543: *fewer than (not less than)

tests/core/common/engines/alpha/steps/events.py
```diff
@@ -527,3 +527,19 @@ def then_the_tool_calls_event_contains_call(
     ]
 
     assert len(matching_tool_calls) > 0, f"No tool call found for {tool_name}"
+
+@step(then, parsers.parse("the number of missing parameters is smaller than {number_of_missing:d}"))
+def then_the_number_of_missing_is_smaller_than(
```
kichanyurd, line 532: LLMs are [not great at counting](https://hackernoon.com/why-cant-ai-count-the-number-of-rs-in-the-word-strawberry).

I think we need to capture the actual tool insights passed to the message composer and validate programmatically that it's fewer than the specified number.

src/parlant/core/tools.py
```diff
@@ -245,25 +255,26 @@ def __init__(
...
 
+    # Note that in this function's arguments ToolParameterOptions is optional (initialized to default if not given)
     async def create_tool(
         self,
         name: str,
         module_path: str,
         description: str,
-        parameters: Mapping[str, ToolParameterDescriptor],
+        parameters: Mapping[
+            str, ToolParameterDescriptor | tuple[ToolParameterDescriptor, ToolParameterOptions]
```
kichanyurd, line 276: What's the reason for introducing the union?

src/parlant/core/engines/alpha/engine.py
```diff
@@ -957,6 +964,20 @@ async def _load_context_variable_value(
...
+        # Precedence 0 is the default, so minimal precedence values start from 1
+        positive_precedence = [
+            m.precedence
+            for m in missing_parameters
+            if m.precedence is not None and m.precedence > 0
+        ]
+        if positive_precedence is None:
+            return missing_parameters
+
+        return [m for m in missing_parameters if m.precedence == min(positive_precedence)]
```
kichanyurd, line 979: Wouldn't this make precedence==1 come before precedence==0?

What would be the issue with this impl?

```python
first_precedence = min(m.precedence for m in missing_parameters)
return [m for m in missing_parameters if m.precedence == first_precedence]
```

src/parlant/core/tools.py
```diff
@@ -234,7 +244,7 @@ class _LocalTool:
     creation_utc: datetime
     module_path: str
     description: str
-    parameters: dict[str, ToolParameterDescriptor]
+    parameters: dict[str, tuple[ToolParameterDescriptor, ToolParameterOptions]]
```
kichanyurd, line 247: Same question here - reason for union?

tests/core/common/engines/alpha/steps/tools.py
```diff
@@ -363,6 +335,133 @@ async def create_tool(
...
+                "last_name",
+                "father_name",
+                "mother_name",
+                "entry_type",
+                "n_entries",
+                "donation_target",
+                "donation_percent",
+            ],
+        },
+        "register_for_sweepstake_2": {
```
kichanyurd, line 402: Let's find a more description name than `_2`?

src/parlant/core/tools.py
mc-kfir: Ease of use:
Local tool parameters now has ToolParameterOptions just like Tool, so create_tool has to be compliant with it.
On the other hand, we wanted to keep the existing local tool tests simple as they were (and not add a lot of empty ToolParameterOptions initializers)

src/parlant/core/tools.py
mc-kfir: It's not the same as above - it's making _LocalTool parameters definition identical to Tool parameters definition

tests/core/stable/engines/alpha/features/baseline/tools.feature
```diff
@@ -308,3 +308,28 @@ Feature: Tools
         Then no tool calls event is emitted
         And a single message event is emitted
         And the message mentions that a date is missing
+
+    Scenario: Message generator understands that multiple parameters are missing and communicates that to the user correctly
```
mc-kfir, line 312: misleading: Message generator understands that multiple parameters

mc-kfir: When multiple parameters are missing, the message generator communicates only the ones with the lowest precedence value (1)

tests/core/stable/engines/alpha/features/baseline/tools.feature
```diff
@@ -308,3 +308,28 @@ Feature: Tools
...
+        And the tool "register_for_sweepstake"
+        And an association between "registering_for_a_sweepstake" and "register_for_sweepstake"
+        And a customer message, "Hi, my first name is Sushi, Please register me for a sweepstake with 3 entries. Ask me right away regarding every missing detail."
+        When processing is triggered
+        Then no tool calls event is emitted
+        And a single message event is emitted
+        And the number of missing parameters is exactly 1
+        And the message mentions last name 
+
+    Scenario: Message generator understands that multiple parameters are missing and communicates that to the user accurately
```
mc-kfir, line 324: When multiple parameters are missing, the message generator communicates only the ones with the lowest precedence value (2)

tests/core/stable/engines/alpha/test_tool_caller.py
```diff
@@ -407,3 +407,75 @@ def available_products_by_category(
     assert isinstance(tool_call.arguments["categories"], str)
     assert "laptops" in tool_call.arguments["categories"]
     assert "peripherals" in tool_call.arguments["categories"]
+
+
+async def test_that_a_tool_from_a_plugin_with_missing_parameters_returns_the_missing_by_precedence(
```
mc-kfir, line 412: test_that_a_tool_from_a_plugin_with_missing_parameters_returns_the_missing_ones_by_precedence

Although ideally this would mirror the scenario string

PR #328, opened 2025-03-27
Feature/mc 644 resolve the issue of data entry into indexeddb when multiple

MenachemBrichta:


src/parlant/api/chat/src/lib/boroadcast-channel.ts
```diff
@@ -0,0 +1,27 @@
+const channel = new BroadcastChannel('active_tabs');
+
+const tabId = Date.now() + '-' + Math.random();
```
kichanyurd, line 3: Rename to `broadcast_channel.ts`

PR #331, opened 2025-03-31
Parlant BDD Test Milestone 1 - Modular API Configuration

Relvox:
In order to make it easier to to configure which API Endpoints will be available from `parlant-server`:

1. Introduced new types to wrap API Configuration Modules:
  ```py
  APIConfigurationStep: TypeAlias = Callable[[FastAPI, Container], AsyncIterator[FastAPI]]

  APIConfigurationSteps: TypeAlias = Sequence[APIConfigurationStep]
  ```
2. Every module has been refactored into it's own configuration function, e.g.:
  ```py
  async def configure_agents_router(app: FastAPI, container: Container) -> AsyncIterator[FastAPI]:
    router = agents.create_router(
      agent_store=container[AgentStore],
      tag_store=container[TagStore],
    )
    app.include_router(router, prefix="/agents")
    yield app

  # and
  async def configure_static_files(app: FastAPI, container: Container) -> AsyncIterator[FastAPI]:
    static_dir = os.path.join(os.path.dirname(__file__), "chat/dist")
    app.mount("/chat", StaticFiles(directory=static_dir, html=True), name="static")

    @app.get("/", include_in_schema=False)
    async def root() -> Response:
      return RedirectResponse("/chat")

    yield app
  ```
3. Made all the steps available from a module level `configuration_steps` field in `app.py`:
  ```py
  configuration_steps: APIConfigurationSteps = [
    configure_middlewares,
    configure_cors_middleware,
    configure_exception_handlers,
    configure_static_files,
    configure_legacy_agents,
    configure_agents_router,
    configure_sessions_router,
    configure_index_router,
    configure_services_router,
    configure_tags_router,
    configure_terms_router,
    configure_customers_router,
    configure_utterances_router,
    configure_context_variables_router,
    configure_guidelines_router,
    configure_logs_router,
  ]
  ```
4. In `server.py`'s `setup_container()` we now load the `Container` with the above `configuration_steps`:
  ```py
  c[APIConfigurationSteps] = configuration_steps
  ```
5. `app.py`'s `create_api_app` now iterates over all the `APIConfigurationSteps` in the `Container`:
  ```py
  async def create_api_app(container: Container) -> ASGIApplication:
    app = FastAPI()
    for configuration_step in container[APIConfigurationSteps]:
      async for configured_app in configuration_step(app, container):
        app = configured_app

    return AppWrapper(app)
  ```


src/parlant/api/app.py
```diff
@@ -147,137 +145,187 @@ async def item_not_found_error_handler(
...
+
+async def configure_logs_router(app: FastAPI, container: Container) -> AsyncIterator[FastAPI]:
+    router = logs.create_router(
+        websocket_logger=container[WebSocketLogger],
     )
+    app.include_router(router)
+    yield app
+
+
+configuration_steps: APIConfigurationSteps = [
```
kichanyurd, line 305: Rename to `default_configuration_steps`

src/parlant/api/app.py
```diff
@@ -147,137 +145,187 @@ async def item_not_found_error_handler(
...
+    configure_context_variables_router,
+    configure_guidelines_router,
+    configure_logs_router,
+]
+
+
+async def create_api_app(container: Container) -> ASGIApplication:
+    app = FastAPI()
+    for configuration_step in container[APIConfigurationSteps]:
+        async for configured_app in configuration_step(app, container):
```
kichanyurd, line 328: There's a bug here. Please add a test that creates an API step something like

```python
    @asynccontextmanager
    async def my_api_step(app: FastAPI, container: Container) -> AsyncIterator[FastAPI]:
        print("Setup")
        yield app
        print("Teardown")
```

...with a couple steps, and ensure that both setup and teardown happen in the right stacking sequence.
Also ensure that you can add steps to the container. I have reason to believe you haven't actually tested that! (Can you guess why?)

P.S. I realized the `APIConfigurationStep` isn't defined right... sorry about that. Instead it should probably be something like:

```python
APIConfigurationStep: TypeAlias = Callable[[FastAPI, Container], AsyncContextManager[FastAPI]]
```

src/parlant/api/app.py
Relvox: rebased and fixed

src/parlant/api/app.py
Relvox: rebased and fixed

Relvox:
I've updated the code and description to reflect the discussions.

Is this what you had in mind? :)

PR #339, opened 2025-04-08
Add heuristics to tool call and add more tests with tools

HadarYosef1:
Main changes:
- Add tests for 2 new tools
- Move tests between stable to unstable
- Add heuristics for rejected tools and optional arguments 

Note:  Few shot can have features and that means that they will be added if they match the curent tool features. If they have no feature we will add them anyway. 

Signed-off-by: Hadar Yosef <hadar@emcie.co>


src/parlant/core/engines/alpha/tool_caller.py
```diff
@@ -601,11 +572,57 @@ def _build_tool_call_inference_prompt(
             props={
                 "service_name": batch[0].service_name,
                 "tool_name": batch[0].tool_name,
+                "tool_calls_for_candidate_tool_json_description": self._format_tool_calls_for_candidate_tool_json_description(
+                    candidate_tool=batch[1], has_reference_tools=bool(reference_tools)
```
mc-dorzo, line 576: can you please also add to props the candidate_tool and has_reference_tools so in future someone (maybe us ..) will use those props to generate different json description

src/parlant/core/engines/alpha/tool_caller.py
```diff
@@ -601,11 +572,57 @@ def _build_tool_call_inference_prompt(
...
+            "applicability_rationale": "<A FEW WORDS THAT EXPLAIN WHETHER, HOW, AND TO WHAT EXTENT THE TOOL NEEDS TO BE CALLED AT THIS POINT>",
+            "applicability_score": <INTEGER FROM 1 TO 10>,
+            "argument_evaluations": [
+                {
+                    "parameter_name": "<PARAMETER NAME>",
+                    "acceptable_source_for_this_argument_according_to_its_tool_definition": "<REAPET THE ACCEPTABLE SOURCE FOR THE ARGUMENT FROM TOOL DEFINITION>",
+                    "evaluate_is_it_provided_by_an_acceptable_source": "<BRIEFLY EVALUATE IF THE SOURCE FOR THE VALUE MATCHES THE ACCEPTABLE SOURCE>",
+                    "evaluate_was_it_already_provided_and_should_it_be_provided_again": "<BRIEFLY EVALUATE IF THE PARAMERWE VALUE WAS PROVIDED AND SHOULD BE PROVIDED AGAIN>",
+                    "evaluate_is_it_potentially_problematic_to_guess_what_the_value_is_if_it_isnt_provided": "<BRIEFLY EVALUATE IF IT'S A PROBLEM TO GUESS THE VALUE>",
+                    "is_optional": <BOOL>,
```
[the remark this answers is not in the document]
HadarYosef1: I can remove it if all parameters are optional. What I thought we will like to do in the future is to iterate over the parameters and show the output based on its attributes. For example add is_optional for this parameter only or add more ARQs if the specific parameter is more tricky to extract or more risky to be assumed or guessed. The problem with that is that we may have more than one tool.

src/parlant/core/engines/alpha/tool_caller.py
```diff
@@ -571,27 +557,12 @@ def _build_tool_call_inference_prompt(
 ```json
 {{
     "last_customer_message": "<REPEAT THE LAST USER MESSAGE IN THE INTERACTION>",
-    "most_recent_customer_inquiry_or_need": "<customer's inquiry or need>",
+    "most_recent_customer_inquiry_or_need": "<COSTUMER"S INQUIRY OR NEED>",
```
kichanyurd, line 560: typo: customer

PR #357, opened 2025-04-28
Improve the behavior of handling invalid tool data

mc-kfir:
* Added invalid data to tool insights: currently refers only to value which is not included in the enum values of a parameter
* Changed tool evaluation ARQ - is_missing was replaced by valid_invalid_or_missing (for the 3 possible states of an argument value)
* Invalid data applies also to optional parameters
* Invalid data and missing data have a common base class of problematic data
* Invalid data is sharing the precedence mechanism with missing data (the precedence is for problematic data)

### DIFF
--- a/src/parlant/core/engines/alpha/tool_calling/single_tool_batch.py
+++ b/src/parlant/core/engines/alpha/tool_calling/single_tool_batch.py
@@ -12,6 +12,7 @@ from parlant.core.engines.alpha.guideline_match import GuidelineMatch
 from parlant.core.engines.alpha.prompt_builder import BuiltInSection, PromptBuilder, SectionStatus
 from parlant.core.engines.alpha.tool_calling.tool_caller import (
     MissingToolData,
+    InvalidToolData,
     ToolCall,
     ToolCallBatch,
     ToolCallBatchResult,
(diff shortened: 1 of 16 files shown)


tests/core/stable/engines/alpha/test_tool_caller.py
```diff
@@ -732,6 +732,153 @@ def register_sweepstake(
...
+            {"message": f"Successfully booked flight to {destination} for passenger {passenger_id}"}
+        )
+
+    conversation_context = [
+        (EventSource.CUSTOMER, "Hi, my nemesis would like to book a one-way flight to Hell"),
+    ]
+
+    interaction_history = create_interaction_history(conversation_context)
+
+    ordinary_guideline_matches = [
```
kichanyurd, line 761: Is this necessary to validate the expected behavior here? If it is, maybe add a small comment to hint at its effect? If it isn't, I think we can remove it.

src/parlant/core/engines/alpha/tool_calling/single_tool_batch.py
```diff
@@ -468,17 +509,17 @@ def _format_tool_calls_for_candidate_tool_json_description(
...
+                    "evaluate_was_it_already_provided_and_should_it_be_provided_again": "<BRIEFLY EVALUATE IF THE PARAMETER VALUE WAS PROVIDED AND SHOULD BE PROVIDED AGAIN>",
                     "evaluate_is_it_potentially_problematic_to_guess_what_the_value_is_if_it_isnt_provided": "<BRIEFLY EVALUATE IF IT'S A PROBLEM TO GUESS THE VALUE>","""
         if optional_arguments:
             result += """
                     "is_optional": <BOOL>,"""
 
         result += """
-                    "is_missing": <BOOL>,
-                    "value_as_string": "<PARAMETER VALUE>"
+                    "valid_invalid_or_missing": <STR: EITHER 'missing', 'invalid' OR 'valid' DEPENDING IF THE VALUE IS MISSING, PROVIDED BUT NOT FOUND IN ENUM LIST, OR PROVIDED AND FOUND IN ENUM LIST (OR DOESN'T HAVE ENUM LIST)>",
```
kichanyurd, line 521: I know this sounds weird but I've encountered inconsistencies with LLM outputs when *we* weren't consistent with putting value descriptions in quotes. Therefore, please add quote around the `<....>` content to avoid such edge cases.  Also note that it seems like you have that last quote but the initial one is missing.

tests/core/stable/engines/alpha/test_tool_caller.py
mc-kfir: Following our conversation, opened an issue (#361) for that

Issue #291, opened 2025-02-19
[Enhancement] Client Test Library


kichanyurd:
# Motivation
We've been using BDD tests internally to test Parlant's engine. This has been really convenient and effective to specify many different user stories and see that agents respond according to their expected behavior. Here's an example:

```gherkin
Scenario: The agent sells pizza in accordance with its defined description
    Given an agent whose job is to sell pizza
    And an empty session
    And a customer message, "Hi"
    And a guideline to do your job when the customer says hello
    When processing is triggered
    Then a single message event is emitted
    And the message contains a direct or indirect invitation to order pizza
```

It would be great if people could create a test suite for their agents in the same fashion. This would make it easy to:
1. Test for regressions in expected behavior when updating the configuration (guidelines, glossary, etc.)
2. Evaluate different LLM providers or custom predictors
3. Assess their agent's consistency
4. Generate many scenarios with AI assistance to test agents at scale

# Solution Proposal
I propose that people could create a **pytest** suite in the following manner:

`my_test.py`
```python
import parlant.test
from pytest_bdd import scenarios

scenarios(
    "onboarding.feature",
    "product_support.feature"
)
```

And here's a couple of examples of the syntax I'm proposing.

`credit_card_scenarios.feature`
**Specifying agents and customers by name**
```gherkin
Scenario: The agent offers a thirsty customer a drink
    Given the Thirst Quencher agent
    And the customer Thirsty Joe
    And the following interaction
      |  By        |  Message                      |
      |  customer  |  Hello                        |
      |  agent     |  Hi, how can I assist you?    |
      |  customer  |  I'm thirsty                  |
    When the agent responds
    Then the response offers a drink of some kind
```

`credit_card_scenarios.feature`
**Specifying agents and customers by ID**
```gherkin
Scenario: The agent offers a thirsty customer a drink
    Given the agent with ID 1tI0lvATV7
    And the customer with ID 4xYgnwzPNf
    And the following interaction
      |  By        |  Message                      |
      |  customer  |  Hello                        |
      |  agent     |  Hi, how can I assist you?    |
      |  customer  |  I'm thirsty                  |
    When the agent responds
    Then the response offers a drink of some kind
```

Here, the step `The response (...)` takes `(...)` and performs an NLP yes/no classification of the assertion with respect to the generated response. In other words, something like:

```python
assert nlp_test(
    context=f"Here's a conversation between {agent} and {customer} [...]\n"
            f"And here's the latest response of the agent: {message}",
[...]

PR #364, opened 2025-05-05
Test API Infrastructure

Relvox:
> I've been out of sync for far too long (my fault). Would appreciate eyes on the code to make sure I haven't strayed too far...

This PR covers milestones 1 and 2 of #291 .

* Refactored endpoint/router registration mechanism and made it customizable.
* added a `--test [opt_module]` flag to start the server in a special test mode.
* added endpoints to manually invoke `guideline_matcher` and `tool_call_inference`.
* added minimal tests for all the above.

> I'm keeping the PR text minimal for now, will expand as details become finalized.


src/parlant/core/engines/alpha/guideline_matcher_test_api.py
mc-dorzo, line 1: move those files under src/parlant/api/ and also the tests under tests/api/

tests/api/test_app_wrapper.py
```diff
@@ -0,0 +1,114 @@
...
+import pytest
+from contextlib import asynccontextmanager
+from fastapi import FastAPI
+from typing import AsyncIterator, List
+
+from parlant.api.app import APIConfigurationSteps, AppWrapper
+from lagom import Container
+
+
+class MockLogger:
```
mc-dorzo, line 25: you can just user container[Logger]

Relvox: can you explain more about what you mean?
I'm using the `MockLogger` to verify which steps get called. Not sure how to do that directly with a `Logger`.
And `container[Logger]` doesn't expose fields to easily check what lines it got.

Relvox:
rebase + tests ran and passed

PR #368, opened 2025-05-06
Change all servers/clients in tests to use random ports + fixes

mc-kfir:
* Changed the test context API class to randomize the port it uses
* Changed run_server wrapper to use randomized port from test context API (which holds the client)
* Reworked run_server termination so it won't leave so many zombie procs
* Changed run_service servrer to use random port
* Changed run_openapi_server to use random port
* Changed error handling of openapi's call_tool to pass exceptions as ToolError (fix for a failed test)
* Added waits for server runners that test they are serving before returning them to the test
* Fixed some more spelling typos

### DIFF
--- a/tests/e2e/test_utilities.py
+++ b/tests/e2e/test_utilities.py
@@ -26,7 +26,9 @@ import sys
 import time
 from typing import Any, AsyncIterator, Iterator, Optional, TypedDict, cast
 
-from tests.test_utilities import SERVER_ADDRESS, SERVER_PORT
+from random import randint
+
+from tests.test_utilities import SERVER_ADDRESS, SERVER_ADDRESS_BASE, SERVER_PORT
 
 
 class _ServiceDTO(TypedDict):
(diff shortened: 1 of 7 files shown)


kichanyurd:
Is the randomization to allow test parallelism? If so, I think it's better to keep track of ports we already used. Even if the range is wide, it's possible to hit the same port again, and it'd be nice to avoid this (admittedly small) probability.

mc-kfir:
> Is the randomization to allow test parallelism? If so, I think it's better to keep track of ports we already used. Even if the range is wide, it's possible to hit the same port again, and it'd be nice to avoid this (admittedly small) probability.

Keeping track of ALL ports of all server kinds on all concurrent tests is quite a hustle and I don't think it's required here because if the port is in use then the port is randomized again (up to 3 retries by default) - so this small probability becomes even more negligible.

tests/test_utilities.py
```diff
@@ -546,12 +548,13 @@ async def _document_loader(doc: BaseDocument) -> Optional[SchematicGenerationRes
 
 @asynccontextmanager
 async def run_service_server(
-    tools: list[ToolEntry],
-    plugin_data: Mapping[str, Any] = {},
+    tools: list[ToolEntry], plugin_data: Mapping[str, Any] = {}, port: int = 0
 ) -> AsyncIterator[PluginServer]:
+    if port == 0:
```
mc-dorzo, line 553: also need to be with function of `available` random port

tests/e2e/test_utilities.py
```diff
@@ -74,9 +76,22 @@ def is_server_responsive(port: int) -> bool:
 def run_server(
     context: ContextOfTest,
     extra_args: list[str] = [],
+    port: int = SERVER_PORT,
+    randomize_port: bool = False,
```
mc-dorzo, line 80: No need for randomize_port. Actually, since we already set the port inside the context, you can remove the port (and port_retries) as well. lets go with that until we discover we need it..

PR #374, opened 2025-05-11
Added choices to insights of missing/invalid tool parameters (if they have enum/choice_provider)

mc-kfir:
* Added choices to ProblematicToolData
* Added a prompt instruction to present the user with the choices when relevant
* Updated relevant existing tests to also check the choices
* Some more spelling corrections


src/parlant/core/engines/alpha/message_generator.py
```diff
@@ -550,7 +551,7 @@ def _build_prompt(
 -------------------------------------
 The following is a description of data that has been provided but are not valid values for their tool parameters in order to run tools.
 The tools would have run, if they only had this data available and there was no missing data.
-You should inform the user about this invalid data: ###
+You should inform the user about this invalid data and if it includes choices then present all of the choices to the user. Here is the invalid data: ###
```
mc-dorzo, line 554: add heuristic

PR #388, opened 2025-05-26
✨ Added Support for Local LLM Services (Ollama / LM Studio)

kunwar-vikrant:
## ✨ Added Support for Local LLM Services (Ollama / LM Studio)

This PR adds support for using local LLM services such as **Ollama**.

### ✅ Tested Setup
- Successfully tested with **ollama** using the `gemma-3` model.

**(Important)Note:** Increase the context length of the model you are using , for me it works for most of it ~14k"

### Switch Models:
- Test another model using environment variable, for eg. export OLLAMA_MODEL="llama3.1:8b"


src/parlant/adapters/nlp/ollama_service.py
```diff
@@ -0,0 +1,349 @@
...
+
+        # Format prompt for Gemma 3
+        formatted_prompt = (
+            f"<start_of_turn>user\n{prompt}\n<end_of_turn>\n<start_of_turn>assistant\n"
+        )
+
+        lmstudio_api_arguments = {k: v for k, v in hints.items() if k in self.supported_ollama_params}
+
+        # Simplified JSON schema
+        json_schema = {
```
kichanyurd, line 169: Hey, looks like you're forcing the schema for the message generator here, whereas in practice this generator is used for all LLM requests - meaning that the schema would be wrong for the rest of them. Did this work when you ran it?

kunwar-vikrant: Hi @kichanyurd , Yeah actually i did need to force the schema here to make this work otherwise it just doesn't responds with expected schema and throws error, i am trying to figure this out but so far i have been able to test this and it works with several local models in lm-studio with context length ~14k.

kichanyurd: Hmm, weird! Parlant uses multiple schemas for its different components (guideline matching, tool calling, message generation, etc. -- actually multiple schemas are supported per each of those components), so it doesn't seem right that it works when you force specifically the message generator's schema on all of them. 

Can you share the error it returns when you don't try to force the schema? We might find a way to fix that in a cleaner way together.

kunwar-vikrant: @kichanyurd , seems like there is some issue with LM Studio's API, anyways i found it better to just fix ollama for now, seems to be working as expected for now, can you also have a look and try it out with some local models if possible, so far i have tried with ollama and qwen, seems to be working as expected. I have removed enforcing schemas as well.

mc-dorzo: Hey @kunwar-vikrant, can you please add some guidelines to your system setup, connect them to tools, and try running it again? The schema must be dynamically inserted—I’m sure the current approach won’t work with the full flow unless that’s handled.

kunwar-vikrant: Hey @mc-dorzo , let me got through this , will update you as soon as possible.

kunwar-vikrant: Hi Guys!,
@kichanyurd @mc-dorzo 
so this is what i recall doing during my testing:
- Installed the package in "editable mode" from the current directory with my changes.
- Installed sentence-transformers ("pip install sentence-transformers")
- Started ollama server running gemma3
- After following the parlant installation guide , exported env variable "export OLLAMA_MODEL=gemma3"
- Executed the server using "parlant-server run --ollama"
- Followed the guide on [Server and Agent Setup ](https://www.parlant.io/docs/tutorial/setup) to create an example agent.

I was able to get valid responses using this guide. Let me know if you guys have an extensive test in mind, i will test the scenario and post the results here using ollama.

iwr-redmond:
This PR should be closed in favor of #496.

PR #389, opened 2025-05-26
Feature/mc 793 fix potential hallucinated ids for guideline matchers

MCBarKar:
2 small changes:
1. In guideline matching prompts, guidelines are now recognized by natural numbers, instead of by their ID. This reduces the chance of the LLM hallucinating a guideline ID.
2. If the guideline matcher inference returns an invalid ID, it will be skipped instead of raising an exception.

A future change would add an extra batch for recovering guideline propositions with invalid IDs

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -6512,4 +6512,4 @@ together = ["together", "torch", "transformers"]
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.10"
-content-hash = "576780bd541187081ff26ced4ea10a0a01c09f46da8c85488342807dfa2b94f6"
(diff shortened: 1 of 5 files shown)


kichanyurd:
@MCBarKar The guideline matcher _used to_ look at guidelines by a running index, but there were issues there, upon which I changed it to use the ID instead. Unfortunately, I can't recall now exactly what they were, but you can `git blame` and probably find it. But for this reason, I'm currently disinclined to bring that approach back.

MCBarKar:
@kichanyurd 
I think I found what you're talking about, is it commit 0dd459f? I didn't find any info as to why we made that change in the commit message, is there anywhere else where we can find it?

I made this PR since we got a report about a case where the LLM hallucinated a guideline ID. I thought that simplifying the IDs would reduce the chance for hallucinations.

PR #408, opened 2025-06-04
Feature/mc-820-fix-incorrect-parsing-of-response-analysis-output-in-testing

HadarYosef1:
Main changes:
1. Fix missing check if guideine previously applied that response analysis result
2. Change the name of GuidelinePreviouslyApplied to AnalyzedGuideline because I think GuidelinePreviouslyApplied sounds like they all applied, when they are not.  

### DIFF
--- a/tests/core/stable/engines/alpha/test_guideline_matcher.py
+++ b/tests/core/stable/engines/alpha/test_guideline_matcher.py
@@ -413,10 +413,10 @@ def create_guideline_by_name(
 
 def update_previously_applied_guidelines(
     context: ContextOfTest,
-    session: Session,
+    session_id: SessionId,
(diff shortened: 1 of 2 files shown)


src/parlant/core/engines/alpha/engine.py
```diff
@@ -989,9 +989,7 @@ async def _add_agent_state(
         )
 
         applied_guideline_ids = [
-            p.guideline.id
-            for p in analysis_result.previously_applied_guidelines
-            if p.is_previously_applied
+            p.guideline.id for p in analysis_result.analyzed_guidelines if p.is_previously_applied
```
mc-dorzo: analysis_result -> result

PR #419, opened 2025-06-09
Feature/mc-792-support-guidelines-with-conditions-that-apply-to-the-agents

HadarYosef1:
Main changes:

- Implement guideline matching for guidelines with condition that apply to the agent:
Detect whether its condition refers to "agent intention". 
If it does, rephrase the condition as "the agent is likely to (do something)". Now it should be clear that we are trying to speculate whether the agent is likely to do something based on the interaction so far. If it's probable, it will be apply. 
* I tried phrasing like "the agent intends to.." but it didn't cover cases when the next agent step is not certain, but we want to consider the guideline if the condition is likely and reasonable, even if not explicitly confirmed.
Made no change in guideline matching prompting. 

- Add tests to test_guideline_matcher, with the proper condition phrasing (the agent is likely to...)

-Add tests to agent intention proposer - check that conditions detect and rephrase correctly, and check that after rewriting the guidelines they are matched.

Change to message generator / utterance selector:
- Add an explanation about guidelines with agent intention to message producers, because don't need to adhere to those guideline if their condition is not eventually applied to the user
- Add tests to BDDs to test how message producers deal with such guidelines, and with several guidelines that override each other.
-   We found that some of those tests are failing, even after adding clarification to the prompt and adding a shot. The scenario that is failing is not really related to agent intention: if we have a guideline when something then do x or y, then a guideline when something more specific then don't do x, we expect it to consider both guidelines and do y. There is more work to be done to make those scenarios pass. For now I moved those tests to unstable. I'll note that these test are not passing in develop in similar rate. 

### DIFF
--- a/src/parlant/core/engines/alpha/guideline_matching/generic/common.py
+++ b/src/parlant/core/engines/alpha/guideline_matching/generic/common.py
@@ -1,0 +1,17 @@
+from dataclasses import dataclass
+from typing import Optional
+
+from parlant.core.guidelines import Guideline
+
+
+@dataclass
+class GuidelineInternalRepresentation:
+    condition: str
+    action: Optional[str]
+
+
+def internal_representation(g: Guideline) -> GuidelineInternalRepresentation:
+    condition = g.metadata.get("internal_condition")
+    if isinstance(condition, str):
+        return GuidelineInternalRepresentation(condition, g.content.action)
+    return GuidelineInternalRepresentation(g.content.condition, g.content.action)
--- a/src/parlant/core/engines/alpha/guideline_matching/generic/common.py
+++ b/src/parlant/core/engines/alpha/guideline_matching/generic/common.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Optional
+from typing import Optional, cast
 
 from parlant.core.guidelines import Guideline
 
@@ -11,7 +11,8 @@ class GuidelineInternalRepresentation:
 
 
 def internal_representation(g: Guideline) -> GuidelineInternalRepresentation:
-    condition = g.metadata.get("agent_intention_condition")
-    if isinstance(condition, str):
-        return GuidelineInternalRepresentation(condition, g.content.action)
-    return GuidelineInternalRepresentation(g.content.condition, g.content.action)
+    action, condition = g.content.action, g.content.condition
+
+    condition = cast(str, g.metadata.get("agent_intention_condition", condition))
+
+    return GuidelineInternalRepresentation(condition, action)
--- a/src/parlant/core/engines/alpha/guideline_matching/generic/common.py
+++ b/src/parlant/core/engines/alpha/guideline_matching/generic/common.py
@@ -13,6 +13,7 @@ class GuidelineInternalRepresentation:
 def internal_representation(g: Guideline) -> GuidelineInternalRepresentation:
     action, condition = g.content.action, g.content.condition
 
-    condition = cast(str, g.metadata.get("agent_intention_condition", condition))
+    if agent_intention_condition := g.metadata.get("agent_intention_condition"):
+        condition = cast(str, agent_intention_condition) or condition
 
     return GuidelineInternalRepresentation(condition, action)
--- a/src/parlant/core/engines/alpha/guideline_matching/generic/common.py
+++ b/src/parlant/core/engines/alpha/guideline_matching/generic/common.py
@@ -11,7 +11,7 @@ class GuidelineInternalRepresentation:
 
 
 def internal_representation(g: Guideline) -> GuidelineInternalRepresentation:
-    condition = g.metadata.get("internal_condition")
+    condition = g.metadata.get("agent_intention_condition")
     if isinstance(condition, str):
         return GuidelineInternalRepresentation(condition, g.content.action)
     return GuidelineInternalRepresentation(g.content.condition, g.content.action)
--- a/src/parlant/core/engines/alpha/guideline_matching/generic/guideline_actionable_batch.py
+++ b/src/parlant/core/engines/alpha/guideline_matching/generic/guideline_actionable_batch.py
@@ -66,6 +66,9 @@ class GenericActionableGuidelineMatchingBatch(GuidelineMatchingBatch):
                 hints={"temperature": 0.15},
             )
 
+        with open("output_intention_checks.txt", "a") as f:
+            f.write(inference.content.model_dump_json(indent=2))
+
         if not inference.content.checks:
             self._logger.warning("Completion:\nNo checks generated! This shouldn't happen.")
         else:
@@ -152,7 +155,7 @@ class GenericActionableGuidelineMatchingBatch(GuidelineMatchingBatch):
         shots: Sequence[GenericActionableGuidelineGuidelineMatchingShot],
     ) -> PromptBuilder:
         guidelines_text = "\n".join(
-            f"{i}) Condition: {g.content.condition}. Action: {g.content.action}"
+            f"{i}) Condition: {g.internal_condition}. Action: {g.content.action}"
             for i, g in self._guidelines.items()
         )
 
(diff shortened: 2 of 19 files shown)


src/parlant/core/guidelines.py
```diff
@@ -56,6 +56,13 @@ def __str__(self) -> str:
     def __hash__(self) -> int:
         return hash(self.id)
 
+    @property
+    def internal_condition(self) -> str:
```
kichanyurd: We might have a lot of these optimizations in the future, and, as they're quite contingent on our current observations, I think we should not treat them as first-class citizens in our architecture (e.g., affecting the API of a core class such as `Guideline`.

Instead, I think the way you started this is already good: keep it in the metadata.

Now, under a new file `guideline_matching/generic/common.py` you can define a helper:

```python
@dataclass
class GuidelineInternalRepresentation:
    condition: str
    action: str

def internal_representation(g: Guideline) -> GuidelineInternalRepresentation:
  ...
```

and in all of the places where you replaced `g.content.condition` with `g.internal_condition`, you can replace that with `internal_representation(g).condition`.

The advantage of this is that the core guideline class remains clean. It's only an internal optimization (which the core class is and shouldn't be aware of by itself), which is evaluated and performed under specialized engine classes (under guideline matching).

Hope that made sense.

MCBarKar:
Documenting the last fixes we talked about here:
- Add the new schema to server.py
- Differentiate agent-intention guidelines from other guidelines in the message producer prompts. Look into editing a few shot there and add a couple of tests

MCBarKar:
Can be merged as far as I'm concerned - there are some further improvements that should be made to the message producer prompts in general. But they can be done separately.

@mc-dorzo

tests/core/stable/engines/alpha/features/baseline/fluid_utterances.feature
```diff
@@ -181,4 +181,78 @@ Feature: Fluid Utterance
         When processing is triggered
         Then a single message event is emitted
         And the message contains the name 'Beef'
-        And the message contains a welcoming back of the customer to the store and asking how the agent could help
\ No newline at end of file
+        And the message contains a welcoming back of the customer to the store and asking how the agent could help
+
+    Scenario: The agent follows a guideline with agent intention (fluid utterance)
+        Given a guideline to do not provide any personal medical information even if you have it when the agent discusses a patient's medical record
```
kichanyurd: @HadarYosef1  @mc-dorzo  We need to change the guidelines in all of these tests from "when the agent does X" to "when you do X", as in all other cases we talk to the agent in the second person — e.g., "When the situation is X, Then do Y" — so we need to stay consistent with the 2nd person case

If we change this and all of the tests still pass, great!

tests/core/stable/engines/alpha/features/baseline/strict_utterances.feature
```diff
@@ -324,4 +324,76 @@ Feature: Strict Utterance
         And an utterance, "I apologize but I don't have this information available"
         When processing is triggered
         Then a single message event is emitted
-        And the message contains the text "Your current balance is $1244 as of today."
\ No newline at end of file
+        And the message contains the text "Your current balance is $1244 as of today."
+
+    Scenario: The agent follows a guideline with agent intention (strict utterance)
```
kichanyurd, line 329: Remember to change to 2nd person in this file as well

src/parlant/core/engines/alpha/guideline_matching/generic/common.py
```diff
@@ -0,0 +1,17 @@
...
+
+@dataclass
+class GuidelineInternalRepresentation:
+    condition: str
+    action: Optional[str]
+
+
+def internal_representation(g: Guideline) -> GuidelineInternalRepresentation:
+    condition = g.metadata.get("agent_intention_condition")
+    if isinstance(condition, str):
```
kichanyurd: `isinstance()` is a non-trivial runtime check - let's instead just use `cast(str, condition)` here

kichanyurd: Actually the whole function can be:

```python
def internal_representation(g: Guideline) -> GuidelineInternalRepresentation:
  action, condition = g.content.action, g.content.condition

  condition = cast(str, g.metadata.get("agent_intention_condition", condition):

  return GuidelineInternalRepresentation(condition, action)
```

src/parlant/core/engines/alpha/guideline_matching/generic/guideline_actionable_batch.py
```diff
@@ -152,7 +153,7 @@ def _build_prompt(
         shots: Sequence[GenericActionableGuidelineGuidelineMatchingShot],
     ) -> PromptBuilder:
         guidelines_text = "\n".join(
-            f"{i}) Condition: {g.content.condition}. Action: {g.content.action}"
+            f"{i}) Condition: {internal_representation(g).condition}. Action: {g.content.action}"
```
kichanyurd: Once we start working with prompts, we should consistently and strictly only use the internal representation, for future compatibility (e.g., if we then change the internal `action` to something, everything should just work — we shouldn't need to go back again and re-support the `action` part TOO manually)

In other words, when building prompts (anywhere), first map all guidelines to their internal rep, and only then format them within strings. Do not directly access non-internal-rep guidelines from prompt strings.

kichanyurd: @HadarYosef1 @MCBarKar @mc-dorzo

src/parlant/core/engines/alpha/message_generator.py
```diff
@@ -301,33 +301,57 @@ def get_guideline_matches_text(
                 {},
             )
         guidelines = []
-
+        agent_intention_guidelines = []
```
kichanyurd, line 312: Let's always try to put an empty line before control statements (for, if, etc.)

src/parlant/core/engines/alpha/message_generator.py
```diff
@@ -301,33 +301,57 @@ def get_guideline_matches_text(
...
         for i, p in enumerate(all_matches, start=1):
-            guideline = f"Guideline #{i}) When {p.guideline.content.condition}, then {p.guideline.content.action}"
-
-            guideline += f"\n    [Priority (1-10): {p.score}; Rationale: {p.rationale}]"
-            guidelines.append(guideline)
+            if p.guideline.content.action:
+                guideline = f"Guideline #{i}) When {p.guideline.content.condition}, then {p.guideline.content.action}"
+                guideline += f"\n   Rationale: {p.rationale}]"
```
kichanyurd, line 317: We lost the priority score here — is that intentional? @HadarYosef1

src/parlant/core/engines/alpha/message_generator.py
```diff
@@ -301,33 +301,57 @@ def get_guideline_matches_text(
...
-        return (
-            """
+        guideline_instruction = """
 When crafting your reply, you must follow the behavioral guidelines provided below, which have been identified as relevant to the current state of the interaction.
-These guidelines were provided by the business you are representing.
-Each guideline includes a priority score to indicate its importance and a rationale for its relevance.
+"""
+        if agent_intention_guidelines_list:
+            guideline_instruction += f"""
+Some guidelines are tied to condition that related to you, the agent. These guidelines are considered relevant because it is likely that you intends to output
```
kichanyurd: Need to pass this part of the prompt through an LLM to fix some grammar issues

src/parlant/core/engines/alpha/utterance_selector.py
```diff
@@ -812,30 +812,52 @@ def _get_guideline_matches_text(
 However, in this case, no special behavioral guidelines were provided.
 """
         guidelines = []
-
+        agent_intention_guidelines = []
```
kichanyurd, line 830: It seems like all of the code here is shared between message generator and utterance selector.

Since it's getting to be non-trivial, let's move it to `PromptBuilder` under `.add_guidelines_for_message_generation()`

src/parlant/core/engines/alpha/message_generator.py
HadarYosef1: Yes, we no longer ask the guideline matchers to return a score for priority.

PR #427, opened 2025-06-19
Feature/mc-861-add-mechanism-for-disambiguating-guidelines

HadarYosef1:
Main changes:

- Add mechanism of guideline disambiguations
- Add test - most of them in test_disambiguation_batch. Add some tests to tests_guideline_matcher to test how guideline matching works when there is an ambiguaity + guidelines that should match from the group. Also Add BDD tests to see that the agent really adhering the clarification guideline + ask for clarification again if the user didn't answer

*I do have 2 tests that failing, but it's on cases that I would rather if it won't include some guidelines in the disambiguation but its not very critical it it does. I may need to change the test so it enable some guidelines to be either include or not include in the disambiguation group and still pass. Anyway, I didnt move them to unstable (Not really saomething to fix here) but I can.

Small implementation details to be aware of:
When listing the members of a disambiguation group head in its metadata, we retrieve them from the list of guidelines passed to the guideline matcher, rather than fetching them from the guideline store. This is done to avoid cases where the designer has linked guidelines from different agents to the same disambiguation head.


tests/core/stable/engines/alpha/test_disambiguation_batch.py
```diff
@@ -0,0 +1,949 @@
...
+        customer,
+        conversation_context,
+        head_condition,
+        is_ambiguous=False,
+        to_disambiguate_guidelines_names=to_disambiguate_guidelines,
+        disambiguating_guideline_names=disambiguating_guidelines,
+    )
+
+
+async def test_that_agent_re_ask_when_customer_didnt_answer_which_option(  # TODO BDD
```
MCBarKar, line 641: Is this here on accident?

tests/core/unstable/engines/alpha/features/user_stories/conversation.feature
```diff
@@ -97,3 +97,36 @@ Feature: Conversation
         Then a single message event is emitted
         And the message contains no direct offer of a 20% discount
 
+Scenario: The agent adhering the clarification guideline when disambiguation needed
```
kichanyurd: `Scenario: The agent adheres to the clarification guideline when disambiguation is needed`

tests/core/unstable/engines/alpha/features/user_stories/conversation.feature
```diff
@@ -97,3 +97,36 @@ Feature: Conversation
...
+        And a disambiguation head "amusement_park" when the customer asks to book a ticket to an amusement ride or attraction, and its not clear which one
+        And a guideline "snake_roller_coaster" is grouped under "amusement_park"
+        And a guideline "turtle_roller_coaster" is grouped under "amusement_park"
+        And a guideline "tiger_Ferris_wheel" is grouped under "amusement_park"
+        And a customer message, "Can I order one ticket to the roller coaster and one ticket to your tiger ferris wheel?"
+        When processing is triggered
+        Then a single message event is emitted
+        And the message contains a suggestion to book snake roller coaster or turtle roller coaster
+
+Scenario: The agent reask for clarification when disambiguation needed and customer hasn't responded
```
kichanyurd: `Scenario: The agent re-asks for clarification when disambiguation is needed and the customer hasn't responded`

tests/core/unstable/engines/alpha/features/user_stories/conversation.feature
```diff
@@ -97,3 +97,36 @@ Feature: Conversation
         Then a single message event is emitted
         And the message contains no direct offer of a 20% discount
 
+Scenario: The agent adhering the clarification guideline when disambiguation needed
+        Given an agent
+        And an empty session
+        And a guideline "snake_roller_coaster" to book it when the customer asks for the snake roller coaster
+        And a guideline "turtle_roller_coaster" to book it when the customer asks for the turtle roller coaster
+        And a guideline "tiger_Ferris_wheel" to book it when the customer asks for the tiger Ferris wheel
+        And a disambiguation head "amusement_park" when the customer asks to book a ticket to an amusement ride or attraction, and its not clear which one
```
kichanyurd: Change step name to `a disambiguation group head` (not just head by itself)

kichanyurd: `And a disambiguation group head "amusement_park" to activate when [...]`

Note the "to activate when..." also

tests/core/unstable/engines/alpha/features/user_stories/conversation.feature
```diff
@@ -97,3 +97,36 @@ Feature: Conversation
...
+        And a guideline "snake_roller_coaster" is grouped under "amusement_park"
+        And a guideline "turtle_roller_coaster" is grouped under "amusement_park"
+        And a guideline "tiger_Ferris_wheel" is grouped under "amusement_park"
+        And a customer message, "Can I order one ticket to the roller coaster and one ticket to your tiger ferris wheel?"
+        When processing is triggered
+        Then a single message event is emitted
+        And the message contains a suggestion to book snake roller coaster or turtle roller coaster
```
kichanyurd: This assertion is problematic. The LLM might interpret it as a logical "X or Y" which would be true if only X applies.

Instead, let's add TWO different assertions (and I suggest using capitals where relevant to be clear):
`And the message contains a suggestion to book the SNAKE roller coaster`
`And the message contains a suggestion to book the TURTLE roller coaster`

tests/sdk/test_guidelines.py
```diff
@@ -0,0 +1,98 @@
...
+
+        self.g1 = await self.agent.create_guideline(condition="Condition 1", action="Action 1")
+        self.g2 = await self.agent.create_guideline(condition="Condition 2", action="Action 2")
+
+        self.priority = await self.g1.prioritize_over(self.g2)
+
+    async def run(self, ctx: Context) -> None:
+        relationship_store = ctx.container[RelationshipStore]
+
+        rel_priority = await relationship_store.read_relationship(id=self.priority.id)
```
kichanyurd: You can reuse the variable name `relationship = ` across all tests instead of `rel_priority`, `rel_entail`, etc. It also reads better.

tests/sdk/test_guidelines.py
```diff
@@ -0,0 +1,98 @@
...
+    async def setup(self, server: p.Server) -> None:
+        self.agent = await server.create_agent(
+            name="Rel Agent",
+            description="Agent for guideline relationships",
+        )
+
+        self.g1 = await self.agent.create_guideline(condition="Condition 1", action="Action 1")
+        self.g2 = await self.agent.create_guideline(condition="Condition 2", action="Action 2")
+
+        self.priority = await self.g1.prioritize_over(self.g2)
```
kichanyurd: `self.relationship` for all tests which have one relationship being tested

tests/sdk/test_guidelines.py
```diff
@@ -0,0 +1,98 @@
...
+
+class Test_that_guideline_entailment_relationship_can_be_created(SDKTest):
+    async def setup(self, server: p.Server) -> None:
+        self.agent = await server.create_agent(
+            name="Rel Agent",
+            description="Agent for guideline relationships",
+        )
+
+        self.g1 = await self.agent.create_guideline(condition="Condition 1", action="Action 1")
+        self.g3 = await self.agent.create_guideline(condition="Condition 3", action="Action 3")
```
kichanyurd: `g3` by mistake? Change all 3s to 2s in this line

tests/sdk/test_guidelines.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+class Test_that_guideline_dependency_relationship_can_be_created(SDKTest):
+    async def setup(self, server: p.Server) -> None:
+        self.agent = await server.create_agent(
+            name="Rel Agent",
+            description="Agent for guideline relationships",
+        )
+
+        self.g2 = await self.agent.create_guideline(condition="Condition 2", action="Action 2")
```
kichanyurd, line 53: 1 and 2, not 2 and 3

tests/sdk/test_guidelines.py
```diff
@@ -0,0 +1,98 @@
...
+    async def run(self, ctx: Context) -> None:
+        assert len(self.relationships) == 2
+
+        for rel in self.relationships:
+            assert rel.kind == GuidelineRelationshipKind.DISAMBIGUATION
+            assert rel.source == self.g1.id
+            assert rel.target in [self.g2.id, self.g3.id]
+
+
+class Test_that_guideline_disambiguate_raises_on_single_target(SDKTest):
```
kichanyurd: `Test_that_attempting_to_disambiguate_a_single_target_raises_an_error`

src/parlant/core/engines/alpha/guideline_matching/guideline_matcher.py
```diff
@@ -233,11 +238,19 @@ async def match_guidelines(
 
         t_end = time.time()
 
+        result_batches = [result.matches for result in batch_results]
+        matches: Sequence[GuidelineMatch] = list(chain.from_iterable(result_batches))
+
+        with self._logger.operation("Transforming guideline matches"):
```
kichanyurd: Use operations mainly when we care about timing things in logs. We can remove it here.

src/parlant/core/engines/alpha/guideline_matching/guideline_match.py
```diff
@@ -33,6 +35,10 @@ class GuidelineMatch:
     score: int
     rationale: str
     guideline_previously_applied: PreviouslyAppliedType = PreviouslyAppliedType.NO
+    metadata: Optional[Mapping[str, JSONSerializable]] = None
```
kichanyurd:
```python
from dataclasses import dataclass, field

metadata: Mapping[str, JSONSerializable] = field(default_factory=dict)
```

src/parlant/core/engines/alpha/guideline_matching/generic_guideline_matching_strategy_resolver.py
```diff
@@ -78,9 +97,16 @@ async def create_matching_batches(
         previously_applied_actionable_batch: list[Guideline] = []
         previously_applied_actionable_customer_dependent_batch: list[Guideline] = []
         actionable: list[Guideline] = []
+        disambiguation_batches: list[Guideline] = []
```
kichanyurd: A few naming changes here:

1. These aren't the batches - these are preliminary sorting groups for later batch creation. Call these `*_guidelines`
2. Change `actionable` to `actionable_guidelines` for consistency
3. `disambiguation_batches` to `disambiguation_guidelines`
4. Remove the concept of `sub_batches` - we only have `batches`

src/parlant/core/engines/alpha/guideline_matching/generic_guideline_matching_strategy_resolver.py
```diff
@@ -283,6 +374,58 @@ def _create_sub_batch_actionable_guideline(
             context=context,
         )
 
+    async def _get_disambiguation_guideline(
+        self,
+        guideline: Guideline,
```
kichanyurd: Rename this one to `candidate`

src/parlant/core/engines/alpha/guideline_matching/generic/disambiguation_batch.py
```diff
@@ -0,0 +1,651 @@
...
+    guideline_head: GuidelineContent
+    expected_result: DisambiguationGuidelineMatchesSchema
+
+
+class GenericDisambiguationGuidelineMatchingBatch(GuidelineMatchingBatch):
+    def __init__(
+        self,
+        logger: Logger,
+        schematic_generator: SchematicGenerator[DisambiguationGuidelineMatchesSchema],
+        guidelines: Sequence[Guideline],
```
kichanyurd: 1. Change to `disambiguation_guideline: Guideline`
2. Add `disambiguation_targets: list[Guideline]`

src/parlant/core/engines/alpha/guideline_matching/generic_guideline_matching_strategy_resolver.py
```diff
@@ -283,6 +374,58 @@ def _create_sub_batch_actionable_guideline(
             context=context,
         )
 
+    async def _get_disambiguation_guideline(
```
kichanyurd: Rename to `_try_get_disambiguation_group(...) -> Optional[tuple[Guideline, list[Guideline]]]`

src/parlant/core/engines/alpha/guideline_matching/generic_guideline_matching_strategy_resolver.py
```diff
@@ -135,6 +169,63 @@ async def create_report_analysis_batches(
...
+        self,
+        matches: Sequence[GuidelineMatch],
+    ) -> Sequence[GuidelineMatch]:
+        result: list[GuidelineMatch] = []
+        guidelines_to_skip: list[GuidelineId] = []
+
+        for m in matches:
+            disambiguation = m.metadata.get("disambiguation") if m.metadata else None
+
+            if disambiguation and "disambiguated_members" in cast(
```
kichanyurd:
```python
if disambiguation := m.metadata.get(...)`:
  targets = disambiguation.get("targets")
```

And note also the structure of the metadata currently is:
```json
{
  "disambiguation": {
    "disambiguation_members": [ ... ]
  }
}
```

And I have a few notes on this:
1. Let's stay consistent with the relationship by saying `disambiguation_targets` (not members)
2. If we're already within the `disambiguation` key, we don't need to repeat it. We can just have:

```json
{
  "disambiguation": {
    "targets": [ ... ]
  }
}
```

src/parlant/core/engines/alpha/guideline_matching/generic_guideline_matching_strategy_resolver.py
```diff
@@ -135,6 +169,63 @@ async def create_report_analysis_batches(
...
+                    )
+                )
+
+                result.append(
+                    GuidelineMatch(
+                        guideline=Guideline(
+                            id=cast(GuidelineId, f"<transient_{generate_id()}>"),
+                            creation_utc=datetime.now(),
+                            content=GuidelineContent(
+                                condition=m.guideline.content.condition,
```
kichanyurd, line 193: Passing the original condition here can be a problem, since, while the GM has enough context (hopefully) to disambiguate (e.g., explicit disambiguation prompt instructions as well as the disambiguation options/targets), the message generator doesn't have this context.

However, the message generator still lists the guidelines as `When X, then Y`.

This could create inconsistencies in the completion, since the MG prompt may not always decide that X actually applies.

I therefore suggest the following change:
1. Make `condition: Optional[str]` in `Guideline`
2. When it is `None`, the guideline should ALWAYS APPLY even without matching. That is, we shouldn't pass it through any LLM. We just say, "Yep, this is a match"
3. In this case, we would be creating a transient always-on guideline by removing the condition. It would be consistent with the above definition in our domain.
4. In the MG (and utterance selector), when encountering a guideline with an empty condition, instead of listing it as `When X, then Y`, just say, `Y`.

Incidentally, this will be really helpful for another feature we need - which is journey-step guidelines. They also won't have a condition associated with them.

PR #428, opened 2025-06-19
Feature/mc 856 enhancement agent capability awareness

MCBarKar:
Implementation for: https://github.com/emcie-co/parlant/issues/423

Added capability awareness - the agent can now be aware of what capabilities (services to be offered) it has, even if it doesn't need to actively use them yet.

Added a section about capabilities the following prompts:
1. Message Generator
2. Utterance Selector
3. All Guideline Matching Batches

Also added tests for all of these modules.

Also added two additional small changes regarding the handling of journeys - 
1. An ARQ to the message generator that points at the next relevant journey step
2. Small clarification regarding returning to earlier journey steps in relevant cases.


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ All notable changes to Parlant will be documented here.
 - Add an improved support for tool parameter types: date, datetime, optional (bug fix) and lists of other allowed types
 - Add integration with MCP servers for tools (currently only tools, limited MCP types support)
 - Refactor UtteranceStore to be implemented as a VectorStore rather than a DocumentStore
+- Introduce Capability
```
kichanyurd: Change to `- Introduce agent capabilities feature`

src/parlant/api/capabilities.py
```diff
@@ -0,0 +1,385 @@
...
+    "tags": ["tag1", "tag2"],
+}
+
+
+class CapabilityDTO(
+    DefaultBaseModel,
+    json_schema_extra={"example": capability_example},
+):
+    """
+    A capability represents a functional feature or skill of the system.
```
kichanyurd: Change to "of the agent" instead of "of the system"

src/parlant/core/engines/alpha/engine.py
```diff
@@ -405,6 +407,9 @@ async def _initialize_response_state(
         # mostly on the current interaction history.
         context.state.glossary_terms.update(await self._load_glossary_terms(context))
 
+        # Load the relevant capabilities.
+        context.state.capabilities = list(await self._load_capabilities(context))
```
kichanyurd: Let's `safe_gather` both  glossary and capability loading

src/parlant/core/capabilities.py
```diff
@@ -0,0 +1,538 @@
...
+                await self._tag_association_collection.delete_one(
+                    filters={"id": {"$eq": tag_assoc["id"]}}
+                )
+
+    @override
+    async def find_relevant_capabilities(
+        self,
+        query: str,
+        available_capabilities: Sequence[Capability],
+        max_capabilities: int = 3,
```
kichanyurd: I think we can remove the default value here... Let everybody explicitly specify it

src/parlant/core/capabilities.py
```diff
@@ -0,0 +1,538 @@
...
+        self,
+        capability_id: CapabilityId,
+    ) -> None: ...
+
+    @abstractmethod
+    async def find_relevant_capabilities(
+        self,
+        query: str,
+        available_capabilities: Sequence[Capability],
+        max_capabilities: int = 3,
```
kichanyurd: Remove default value

src/parlant/core/capabilities.py
```diff
@@ -0,0 +1,538 @@
...
+    ) -> Sequence[Capability]:
+        if not available_capabilities:
+            return []
+
+        n_result = sum(
```
kichanyurd, line 475: This is a brilliant approach - but we must document right above it how and why this approach works (compared to just getting ALL the documents and sorting manually)

src/parlant/core/entity_cq.py
```diff
@@ -165,6 +170,43 @@ async def find_guideline_tool_associations(
...
+
+        result = await self._capability_store.find_relevant_capabilities(
+            query,
+            list(all_capabilities),
+            max_capabilities=3,
+        )
+
+        t_end = time.time()
+
+        logger.info(f"Find time {t_end - t_start}")
```
kichanyurd: Remove log + timestamp variables

src/parlant/core/engines/alpha/prompt_builder.py
```diff
@@ -321,6 +323,87 @@ def add_staged_events(
...
+                name=BuiltInSection.CAPABILITIES,
+                template="""
+Below are the capabilities available to you as an agent.
+You may inform the customer that you can assist them using these capabilities.
+If you choose to use any of them, additional details will be provided in your next response.
+Always prefer adhering to guidelines and relevant journey steps, before offering capabilities - only offer capabilities if you have no other instruction that's relevant for the current stage of the interaction.
+Be proactive and offer the most relevant capabilities—but only if they are likely to move the conversation forward.
+If multiple capabilities are appropriate, aim to present them all to the customer.
+If none of the capabilities address the current request of the customer - DO NOT MENTION THEM.
+When providing your full response, list offered capabilities under the "offered_services" key, and not under "factual_information_provided".
```
kichanyurd: @MCBarKar @mc-dorzo This line, referencing "offered_services" and "factual_information_provided" applies only to the heavy-duty message generator - not to the utterance selector. As such, it breaks an abstraction layer here.

Can we remove it? If not, let's specifically add this comment somewhere in the message generator file itself when building the prompt - not in the prompt builder itself.

src/parlant/core/engines/alpha/prompt_builder.py
```diff
@@ -321,6 +323,87 @@ def add_staged_events(
...
+###
+""",
+                props={"capabilities_string": capabilities_string},
+                status=SectionStatus.ACTIVE,
+            )
+        else:
+            self.add_section(
+                name=BuiltInSection.CAPABILITIES,
+                template="""
+When evaluating guidelines, you may sometimes be given capabilities to assist the customer beyond those dictated through guidelines and journeys.
```
kichanyurd, line 369: Do we have any ARQs in our completions that directly refer to capabilities? If not, I don't think we need to mention them in the prompt at all. So the entire `else:` section can just not add a section at all.

src/parlant/core/entity_cq.py
```diff
@@ -13,9 +13,11 @@
 # limitations under the License.
 
 from itertools import chain
+import time
```
kichanyurd: Remove import

src/parlant/core/engines/alpha/prompt_builder.py
kichanyurd: Another option is to pass another parameter `extra_instructions: list[str]` to `add_capabilities_for_message_generation`, so that only in `message_generator.py` we add it.

```python
def add_capabilities_for_message_generation(
        self,
        capabilities: Sequence[Capability],
        extra_instructions: list[str] = [],
    ) -> PromptBuilder:
```

In `message_generator.py`:
```python
builder.add_capabilities(
  capabilities,
  extra_instructions=['When providing your full response, list offered capabilities under the "offered_services" key, and not under "factual_information_provided".']
)
```

PR #476, opened 2025-08-06
Feature/mc-1058-support-invalid-data-in-overlapping-tools

HadarYosef1:
Small implementaion detail - I entered the check of valid data to the if of is_applicable because in overlapping we check is applicable before we evaluate the calls and not for each call  (do it per tool) , and calls and parameter_name can be None. We can discuss if its a good implementation but I think its ok for now 


MCBarKar:
@mc-dorzo @kichanyurd 
Good to go as far as I'm concerned. Just rebased it from develop.

Issue #431, opened 2025-06-22
[Bug] Deal with list arguments that are provided by LLM without brackets


mc-kfir:
# Description
As some kind of hallucination - LLM may rarely provide list arguments as comma-separated elements **without the required brackets** wrapping them.

Example (happened in test_that_a_tool_with_a_parameter_attached_to_a_choice_provider_gets_the_tool_context):
![Image](https://github.com/user-attachments/assets/a23e6b7e-7afc-4860-bc79-ebfecb3724e1)

# How to Reproduce
This bug is hard to preproduce as it depends on non-deterministic LLM behavior.

# Expected Behavior
At this place in code, the arguments types are known, so we can safely split the list even without the brackets, as long as the list delimiter (comma) is correct.

# Environment
 - OS:  iOS
 - Python version 3.10.17
 - Parlant version

# Discussion

mc-kfir:
A very similar bug may also happen in `cast_tool_argument ` (tools.py) and `prepare_tool_arguments ` (mcp_service.py) - which will also be fixed to prevent a similar future bug.

iwr-redmond:
> As some kind of hallucination

That isn't the only possible explanation. Another is poorly written or conflicting guidelines that undermine the LLM's abilities. An extreme example of this in-context misalignment was [recently observed](https://theconversation.com/groks-white-genocide-responses-show-how-generative-ai-can-be-weaponized-257880) in Grok, but [many](https://arxiv.org/abs/2506.02739) [other](https://arxiv.org/abs/2506.06998) [examples](https://arxiv.org/abs/2506.13206) [exist](https://arxiv.org/abs/2506.14948) and are probably quite common.

If it is assumed that Small Language Models are likely to produce erroneous outputs more often, then finalizing #388 will make it easier to troubleshoot the source. It is possible that using a decent local model, for example Qwen3-32B @ Q8_0, may work just as well as GPT-4o, in which case misalignment may prove to be a more helpful conceptual framework for understanding why the guidelines aren't being followed at all times. OTOH, the local model may error out far more often, in which case model size will become a KPI again and looking at [GPT-4.1](https://cookbook.openai.com/examples/gpt4-1_prompting_guide) could prove fruitful.

[...]

PR #496, opened 2025-08-19
Support for local models with ollama

Agam1997:
This PR adds support to using local models via ollama 

Defaults to gemma3:4b-it-qat 

Recommended to pull models before usage:

ollama pull nomic-embed-text:latest
ollama pull gemma3:4b-it-qat

Ready for initial review

@kichanyurd @mc-dorzo


Agam1997:
This PR also solves for this issue: https://github.com/emcie-co/parlant/issues/246 

An identical PR also exists: https://github.com/emcie-co/parlant/pull/388

src/parlant/adapters/nlp/ollama_service.py
```diff
@@ -0,0 +1,614 @@
...
+        elif "8b" in self.model_name.lower():
+            return 8192
+        elif "12b" in self.model_name.lower() or "70b" in self.model_name.lower():
+            return 16384
+        elif "27b" in self.model_name.lower() or "405b" in self.model_name.lower():
+            return 32768
+        else:
+            return 4096
+    
+    # @TODO: This function isn't working properly and doesnt return correct model names - debug and fix
```
kichanyurd: Is this still a TODO? Or is it working now and ready to go?

src/parlant/adapters/nlp/ollama_service.py
```diff
@@ -0,0 +1,614 @@
...
+            model_name=self.embedding_model
+        )
+    
+    @override
+    async def get_moderation_service(self) -> ModerationService:
+        """Get a moderation service (using no moderation for local models)."""
+        return NoModeration()
+
+
+def create_ollama_service(
```
kichanyurd: Any chance we can also get rid of this function and use env vars instead?

src/parlant/adapters/nlp/ollama_service.py
Agam1997: Work in progress still..

src/parlant/adapters/nlp/ollama_service.py
Agam1997: Yeah.. will make it env based..

Agam1997: fixed

src/parlant/adapters/nlp/ollama_service.py
Agam1997: fixed model ensureness

Agam1997:
Ready to merge

Agam1997:
Hey,
I have removed _ensure_embedding_model_exists and _ensure_model_exists and created a separate handler, which will throw SDK error if the passed model is not available via the base url Ollama instance so that everything is consistent.

docilio:
Hopefully, it will be merged

iwr-redmond:
Is it a good idea to hard-code the generation models without having an override option like `OLLAMA_EMBEDDING_MODEL`?

Agam1997:
> Is it a good idea to hard-code the generation models without having an override option like `OLLAMA_EMBEDDING_MODEL`?

Please check, it supports OLLAMA_EMBEDDING_MODEL as env

iwr-redmond:
Thanks, I saw that. I was asking about the main model.

Agam1997:
> Thanks, I saw that. I was asking about the main model.

Will make a custom generic SchematicGenerator class which would support all types of models .. current implementation follows how other nlp services are made

iwr-redmond:
Perfect, thanks. Preconfigured models are ideal for reproducibility, but an option to override these defaults would be a great addition - especially as local models are evolving so rapidly.

Agam1997:
@iwr-redmond done. 

can you please review once, i tried with llama3.2:3b and it works

iwr-redmond:
Perfect

iwr-redmond:
@kichanyurd:

The existing [LiteLLM adapter](https://github.com/emcie-co/parlant/blob/5e4b530e3b2082939d7a2b0da4eda7e0a44985f1/src/parlant/adapters/nlp/litellm_service.py) can [already](https://docs.litellm.ai/docs/providers/ollama) be used to run Ollama models, as can the default OpenAI provider using [aliases](https://github.com/emcie-co/parlant/issues/359#issuecomment-3146714152).

This PR removes the need to utilize such backdoor methods, creating a more reliable baseline for local experimentation and private deployment, such as in Docker containers. If @Agam1997 is correct that a tiny 4B model can perform well enough out of the box, that will go some way to answering the questions posed about model alignment versus model size in [#431 (comment)](https://github.com/emcie-co/parlant/issues/431#issuecomment-2994125431).

Agam1997:
I found that llama3.2:3b actually performed better than the gemma3:4b 

Although it took a lot of time to respond (which i suppose is a hardware capacity issue) 

Llama 3.2 3b was able to run the healthcare agent example and also told me about the insurances which i can get

iwr-redmond:
Very interesting indeed.

LLMs [prefer](https://arxiv.org/abs/2404.10198v1) to use their priors, so one of the issues with a city-block sized model like GPT4o is that it can resist runtime modifications. SLMs are less likely to put up this type of resistance because they have fewer priors, which means that the main qualifications for Parlant compatibility will be function calling and context adherence rather than size.

The downside to local LLMs is that different models will be suitable for different customers; a fair amount of the deployment/consultancy cost will be finding a model that works well with Parlant and responds appropriately to the tailored guidelines that are being written for the particular use case - for example maintaining cohesion as the context size increases.

pyproject.toml
```diff
@@ -101,6 +101,7 @@ deepseek = ["torch", "transformers"]
 gemini = ["google-genai", "google-api-core", "torch"]
 
 vertex = ["google-genai", "google-api-core", "google-auth", "torch", "anthropic", "transformers"]
+ollama = ["ollama"]
```
mc-dorzo, line 107: This is great, you put it under the extras, but the actual package with the version is missing. Please add it as an optional package. Take a look at other packages.

Agam1997: Hi, thanks for reviewing.

Have resolved this in latest commit.

kichanyurd:
@iwr-redmond @Agam1997 @docilio 

Merged. Thanks for your effort in coding and reviewing this. Enjoy!

(We'll release this as part of 3.0.2 quite soon)

iwr-redmond:
@Agam1997 has been the one doing a land-office business this time around.

docilio:
Well done everyone!! @Agam1997 you were the master, happy to help out

PR #505, opened 2025-08-21
Small fix to support litellm via the SDK

Agam1997:
@kichanyurd 

Kindly review - Added litellm just like how we did for gemini, also added verify_environment

### DIFF
--- a/src/parlant/adapters/nlp/litellm_service.py
+++ b/src/parlant/adapters/nlp/litellm_service.py
@@ -199,18 +199,19 @@ class LiteLLMService(NLPService):
     def verify_environment() -> str | None:
         """Returns an error message if the environment is not set up correctly."""
 
-        if not os.environ["LITELLM_PROVIDER_MODEL_NAME"]:
+        if not os.environ.get("LITELLM_PROVIDER_MODEL_NAME"):
             return """\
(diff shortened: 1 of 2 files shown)


src/parlant/adapters/nlp/litellm_service.py
```diff
@@ -194,6 +194,23 @@ def max_tokens(self) -> int:
 
 
 class LiteLLMService(NLPService):
+    
+    @staticmethod
+    def verify_environment() -> str | None:
+        """Returns an error message if the environment is not set up correctly."""
+
+        if not os.environ["LITELLM_PROVIDER_MODEL_NAME"]:
```
mc-dorzo: Good work man, just please fix a KeyError bug here, since if there is no `"LITELLM_PROVIDER_MODEL_NAME" in os.environ` than it will raise error.

write either (I prefer the first) `KEY not in os.environ` or `os.environ.get(KEY)` (do to the other env vars obviously)

Agam1997: Oh my bad, i see the problem now

Agam1997: Resolved

Agam1997:
@mc-dorzo Done with the changes

PR #517, opened 2025-08-26
Modified LiteLLM nlp class to optionally accept Provider URL for self hosted LLMs

chibexme:


iwr-redmond:
According to the [documentation](https://docs.litellm.ai/docs/providers/openai_compatible), there are three standard fields for completions and embeddings:

- model
- api_key
- api_base

In the absence of the `api_base` parameter, a LiteLLM proxy server appears to be required when addressing a locally hosted model.

@chibexme, it would be helpful to extend this PR to allow for the selection of two models, an LLM and an embedding model.

cc @Agam1997

chibexme:
@iwr-redmond Thanks for your observation, I will work on it.

kichanyurd:
@chibexme Hey mate, please LMK when the changes are made and this is ready to go as far as you can see. No pressure, though.

Issue #522, opened 2025-08-28
[Bug] OLLAMA_EMBEDDING_MODEL not used in


shaunabanana:
# Description
OLLAMA_EMBEDDING_MODEL is not used when calling `agent.create_term()`.

# How to Reproduce
1. Install the latest version (3.0.2)
2. Modify the `examples/healthcare.py` line 166 to be `async with p.Server(nlp_service=p.NLPServices.ollama) as server:`
3. Set OLLAMA_EMBEDDING_MODEL to some model other than the default `nomic-embed-text`.
4. Run `healthcare.py`

The following error:
```
Traceback (most recent call last):
  File "[Project path]/.venv/lib/python3.12/site-packages/parlant/adapters/nlp/ollama_service.py", line 480, in embed
    response = await self._client.embed(model=self.model_name, input=texts)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "[Project path]/.venv/lib/python3.12/site-packages/ollama/_client.py", line 879, in embed
    return await self._request(
           ^^^^^^^^^^^^^^^^^^^^
  File "[Project path]/.venv/lib/python3.12/site-packages/ollama/_client.py", line 692, in _request
    return cls(**(await self._request_raw(*args, **kwargs)).json())
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "[Project path]/.venv/lib/python3.12/site-packages/ollama/_client.py", line 636, in _request_raw
    raise ResponseError(e.response.text, e.response.status_code) from None
ollama._types.ResponseError: model "nomic-embed-text" not found, try pulling it first (status code: 404)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "[Project path]/healthcare.py", line 214, in <module>
    asyncio.run(main())
  File "[Home]/.local/share/uv/python/cpython-3.12.11-macos-aarch64-none/lib/python3.12/asyncio/runners.py", line 195, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "[Home]/.local/share/uv/python/cpython-3.12.11-macos-aarch64-none/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #523, opened 2025-08-28
Fix for #522

Agam1997:
Hi,

This PR is a fix for https://github.com/emcie-co/parlant/issues/522 

The problem was in the OllamaEmbedder constructor - i was putting in base_url and model_name as separate parameters, but the dependency injection system doesn't work that way

Also, updated the docs that embedding models which can produce 512, 768 and 1024 vectors are supported.

### DIFF
--- a/src/parlant/adapters/nlp/ollama_service.py
+++ b/src/parlant/adapters/nlp/ollama_service.py
@@ -418,15 +418,13 @@ class OllamaEmbedder(Embedder):
 
     def __init__(
         self,
-        logger: Logger,
-        base_url: str = "http://localhost:11434",
-        model_name: str = "nomic-embed-text",
+        logger: Logger
     ):
-        self.model_name = model_name
-        self.base_url = base_url.rstrip("/")
+        self.model_name = os.environ.get("OLLAMA_EMBEDDING_MODEL", "nomic-embed-text")
+        self.base_url = os.environ.get("OLLAMA_BASE_URL", "http://localhost:11434").rstrip("/")
         self._logger = logger
-        self._tokenizer = OllamaEstimatingTokenizer(model_name)
-        self._client = ollama.AsyncClient(host=base_url)
+        self._tokenizer = OllamaEstimatingTokenizer(self.model_name)
+        self._client = ollama.AsyncClient(host=self.base_url)
 
     @property
     @override
@@ -450,6 +448,8 @@ class OllamaEmbedder(Embedder):
             return 768
         elif "mxbai" in self.model_name.lower():
             return 512
+        elif "bge" in self.model_name.lower():
+            return 1024
         else:
             return 768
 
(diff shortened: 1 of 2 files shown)


iwr-redmond:
Nice catch, @Agam1997 

Could you please add an environment variable to override the dimensions if desired for a custom embedding model?

src/parlant/adapters/nlp/ollama_service.py
```diff
@@ -418,15 +418,13 @@ class OllamaEmbedder(Embedder):
...
     def __init__(
         self,
-        logger: Logger,
-        base_url: str = "http://localhost:11434",
-        model_name: str = "nomic-embed-text",
+        logger: Logger
     ):
-        self.model_name = model_name
-        self.base_url = base_url.rstrip("/")
+        self.model_name = os.environ.get("OLLAMA_EMBEDDING_MODEL", "nomic-embed-text")
```
mc-dorzo: This breaks consistency. In CustomOllamaSchematicGenerator, you get the model name and base URL (based on the environment variables, but the embedder does not. I think what was before was fine.
Also, please ensure you have the Ruff extension installed and run mypy before. I see that there are a few mypy errors in the file currently.

Agam1997: Will take a look.

I have ruff and mypy, the linter however always gives me an error:

[44] stderr:
mypy: "parlant/src/parlant/core/engines/types.py" shadows library module "types"
note: A user-defined top-level module with name "types" is not supported

[44] Error running mypy in /Users/agamdubey/open-source: mypy failed with error: "mypy: "parlant/src/parlant/core/engines/types.py" shadows library module "types"". See Output panel for details.

Agam1997: @mc-dorzo pushed a fix as we discussed

src/parlant/adapters/nlp/ollama_service.py
```diff
@@ -489,6 +475,64 @@ async def embed(
...
+    def max_tokens(self) -> int:
+        return 8192
+
+    @property
+    def dimensions(self) -> int:
+        return 1024
+
+class OllamaBgeM3EmbeddingLarge(OllamaEmbedder):
+    def __init__(self, logger: Logger) -> None:
+        self.base_url = os.environ.get("OLLAMA_BASE_URL", "http://localhost:11434")
```
mc-dorzo: Since base_url is used in the OllamaEmbedder generally, it is sufficient to populate this variable in the OllamaEmbedder once.

Agam1997: My bad. Thanks dorzo for reviewing! :)

Agam1997:
@iwr-redmond added a new OllamaCustomEmbedding class which will take a vector size from ENV for custom / other models -> falls back to it if there's no nomic, mxbai or bge-m3 embedding models

iwr-redmond:
Thanks @Agam1997!

PR #528, opened 2025-09-01
Feature/mc 1195 enhancement add modules layer

mc-dorzo:


src/parlant/bin/server.py
```diff
@@ -551,8 +550,21 @@ async def setup_container() -> AsyncIterator[Container]:
 
     _define_singleton(c, Engine, AlphaEngine)
 
-    # The application object requires the actual container
-    c[Application] = lambda rc: Application(rc)
+    c[AgentModule] = Singleton(AgentModule)
```
kichanyurd: Please use `_define_singleton` as above

PR #529, opened 2025-09-01
Feature/snowflake cortex adapter

Taooo-habitus:
Hi all!

First off — thanks for this amazing tool! 🙌  
One concern I’ve had is sending data outside of Snowflake, so I built an adapter to seamlessly integrate **Snowflake Cortex** into Parlant.

## Changes
1. **Added a Snowflake Cortex adapter**  

2. **Fixed a few mypy errors**  
   - Added targeted ignores for missing provider stubs  
   - Cleaned up return types flagged as `Any`  

## Next Steps
- Documentation is on the way...


mypy.ini
```diff
@@ -7,4 +7,20 @@ mypy_path = src
 files = src, tests
 disable_error_code = type-abstract
 exclude = scripts
-plugins = pydantic.mypy
```
mc-dorzo: We prefer not to apply a blanket ignore_missing_imports for entire modules. Instead, we try to handle the case of missing imports on a case-by-case basis to avoid unintentionally hiding real issues.

src/parlant/adapters/nlp/snowflake_cortex_service.py
```diff
@@ -0,0 +1,434 @@
...
+        try:
+            self.encoding = tiktoken.encoding_for_model(self.model_name)
+        except Exception:
+            self.encoding = tiktoken.get_encoding("cl100k_base")
+
+    @override
+    async def estimate_token_count(self, prompt: str) -> int:
+        """Estimate token count for a prompt.
+
+        Args:
```
mc-dorzo: Some of these docstrings are redundant. Please remove class and function comments that don’t add value beyond what the code already makes obvious. Keep documentation only where it clarifies behavior that isn’t trivial to understand from the code itself.

src/parlant/adapters/nlp/snowflake_cortex_service.py
```diff
@@ -0,0 +1,434 @@
...
+        }
+
+        for k in self._provider_params:
+            if k in hints:
+                payload[k] = hints[k]
+
+        # Strict path: provider-enforced JSON schema
+        if hints.get("strict", False):
+            try:
+                payload["response_format"] = {
```
mc-dorzo, line 137: default values for by_alias is True, and ref_template="#/$defs/{model}", so we can just write schema.model_json_schema()

Taooo-habitus:
Thanks for the comments! It's now ready for review :)

kichanyurd:
@Taooo-habitus will review and merge soon mate, thank you for this. Would also love to know more about your use case: is it a project related to people with special needs?

Taooo-habitus:
> @Taooo-habitus will review and merge soon mate, thank you for this. Would also love to know more about your use case: is it a project related to people with special needs?

Hi @kichanyurd, that’d be great, thanks!

Yes, our business is focused on helping people with special needs. This particular project involves handling a lot of sensitive data, but it’s primarily aimed at supporting our internal business operations rather than being a direct customer-facing tool. 😊

kichanyurd:
Hey mate @Taooo-habitus we're ready to merge your contribution. Just one quick request.
It seems like you used different formatting rules and, as a result, made a lot of changes to `sdk.py` that aren't to do with your change.

Could you please reset those formatting changes and keep only the addition of the NLP service?

Taooo-habitus:
@kichanyurd

Done :)

kichanyurd:
Great work mate, appreciate that. Merging. @Taooo-habitus

PR #536, opened 2025-09-04
Explicitly output the MongoDB session-related library name in the error message.

sapsaldog:
I tried to use MongoDB as a session store in Parlant, but I got this message:

```
parlant.sdk.SDKError: MongoDB requires an additional package to be installed. Please install parlant[mongo] to use MongoDB.
MongoDB requires an additional package to be installed. Please install parlant[mongo] to use MongoDB.
```
I tried several times with like

```
uv add parlant[mongo]
```

and also searched on Google and Discord, but eventually I checked the source code and found out that the required library is actually pymongo.

I think it would be better if others didn’t have to go through the same experience.

### DIFF
--- a/src/parlant/sdk.py
+++ b/src/parlant/sdk.py
@@ -2768,7 +2768,7 @@ class Server:
                 if importlib.util.find_spec("pymongo") is None:
                     raise SDKError(
                         "MongoDB requires an additional package to be installed. "
-                        "Please install parlant[mongo] to use MongoDB."
+                        "Please install pymongo to use MongoDB."
                     )
 
                 from pymongo import AsyncMongoClient


kichanyurd:
Hmm there might be a problem with uv integration.

Can you please share the output of installing parlant[mongo]? Because in principle it should work.

sapsaldog:
Sure here is the output.

I am using macOS 15.6.1, 
uv 0.8.8 (Homebrew 2025-08-09)

Thanks!

```
sapsaldog@AKX-J47QCXMLWP parlant-sample % uv add parlant[mongo]
zsh: no matches found: parlant[mongo]
sapsaldog@AKX-J47QCXMLWP parlant-sample % uv add pymongo
Resolved 169 packages in 8.71s
Prepared 1 package in 2.70s
Installed 1 package in 16ms
 + pymongo==4.14.1
```

sapsaldog:
It turned out the issue wasn’t with uv or parlant, but with zsh globbing. Since zsh interprets square brackets as a pattern, uv add parlant[mongo] failed with “no matches found.” Wrapping it in quotes fixes it:

```
uv add 'parlant[mongo]'
```

sapsaldog:
That’s a tricky issue. It seems that zsh users might need some extra information.

I’ll leave it up to the reviewers to decide whether to simply close this issue or provide additional guidance.

Thank you for taking the time.

kichanyurd:
I think we'll just update the docs to be clearer on this point, as this is standard pip syntax. Thanks for pointing it out and doing the work!

kichanyurd:
@sapsaldog Updated our docs to include the quotation marks. Once again, good catch, and thanks for putting your time into this.

PR #540, opened 2025-09-07
fix: Handle invalid cron expressions in context variable freshness check

Mirza-Samad-Ahmed-Baig:
Adds a try...except block to handle the CroniterBadCronError exception. If the exception is caught, the error is logged and the existing value is returned. This prevents the application from crashing and ensures that the conversation can continue.


mc-dorzo:
This is a great observation. I think the issue is not in the engine runtime but rather that we may end up with an incorrect freshness value. We should validate it at creation time instead.

Please take a look at the `validate_freshness_rules` function in the API codebase. We need to implement the same validation behavior in the SDK. Would you like to take this on?

Mirza-Samad-Ahmed-Baig:
Surely, it is good one.

tests/core/test_context_variables.py
```diff
@@ -0,0 +1,40 @@
+import pytest
```
kichanyurd: Good on you mate for adding the test file!

However, our current convention is to test stores via the API level (I know, I know... we'll get there).

So could you please add the relevant test (following the same coding conventions in the file) under the `tests/api` dir?

commit_message.txt
```diff
@@ -0,0 +1 @@
+fix: Handle invalid cron expressions in context variable freshness check
```
kichanyurd: Need to remove this file.

dco_remediation.txt
```diff
@@ -0,0 +1,7 @@
+Samad <Mirzasamadahmedbaig@gmail.com>
```
kichanyurd: Also need to remove this file.

kichanyurd:
@Mirza-Samad-Ahmed-Baig are you still on this PR? It's a very good touch-up and we'd love to credit you with it through this contribution. Otherwise we'll need to close it and implement it ourselves. Thanks!

Mirza-Samad-Ahmed-Baig:
Yeah Sure. Let's do it

Mirza-Samad-Ahmed-Baig:
Thank you, done with changes.

PR #546, opened 2025-09-10
Added reference documentation

onel:
First batch of files with reference docs.

Looking for feedback on style / helpfulness, etc

### DIFF
--- a/src/parlant/api/chat/src/components/session-list/session-list-item/session-list-item.tsx
+++ b/src/parlant/api/chat/src/components/session-list/session-list-item/session-list-item.tsx
@@ -17,6 +17,18 @@ import {copy, exportToCsv, getIndexedItemsFromIndexedDB} from '@/lib/utils';
 import Avatar from '@/components/avatar/avatar';
 import CopyText from '@/components/ui/custom/copy-text';
 
+/**
+ * Props interface for the SessionListItem component
+ * @interface Props
+ * @property {SessionInterface} session - The session data to display
+ * @property {boolean} [disabled] - Whether the component is disabled
+ * @property {boolean} [isSelected] - Whether this session is currently selected
+ * @property {string | null} [editingTitle] - The ID of the session currently being edited
+ * @property {Dispatch<SetStateAction<string | null>>} [setEditingTitle] - Function to set the editing title state
+ * @property {() => void} [refetch] - Function to refetch session data
+ * @property {number} [tabIndex] - Tab index for keyboard navigation
+ * @property {ClassNameValue} [className] - Additional CSS classes
+ */
 interface Props {
 	session: SessionInterface;
 	disabled?: boolean;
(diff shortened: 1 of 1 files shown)


mc-dorzo:
Hey @onel ! 

Thanks for the PR. I’ve tagged our frontend engineer to review it.

src/parlant/core/engines/alpha/guideline_matching/generic/generic_guideline_matching_strategy.py
```diff
@@ -93,6 +101,24 @@ def __init__(
         journey_step_selection_schematic_generator: SchematicGenerator[JourneyNodeSelectionSchema],
         response_analysis_schematic_generator: SchematicGenerator[GenericResponseAnalysisSchema],
     ) -> None:
+        """
```
mc-dorzo, line 104: I don’t believe this is crucial to mention, since all our code is based on the assumption that the arguments are implicit from their names and annotations, and therefore lacks explicit documentation in functions.

mc-dorzo:
Hey @onel !
First of all, I really appreciate the initiative here 🙏

After reviewing, I have a concern: most of our functions are not documented in a way that provides meaningful insights into their actual behavior. For example, in the attached image, the documentation doesn’t really explain what “response analysis” means or how the batching works. At the same time, adding this style of inline documentation may create maintenance challenges—for instance, if batch behavior changes or new behavior is added, the function-level comments may quickly become outdated.

That’s why we intentionally focused our documentation effort in alpha.engine.py, where we describe the more non-trivial code paths. Before we move into documenting specific files, I’d love to understand why this file in particular (the generic guideline matcher) was chosen. From there, we can align on what information developers would actually find valuable to see explained here.
image

onel:
@mc-dorzo I appreciate the feedback and don't feel bad, this is helpful.
You comments make sense. because it helps me understand the documentation style for the project.

Should we avoid inline comments altogether or only add them to complex methods?
If the methods have typing information we should not duplicate that by also adding the `Args:` section?

I agree that the description should be a bit more detailed about what's happening inside the method.

I'm not worried about things going out of date because these docs can be regenerated on code change.
As for the file chosen, I looked at the files that were lacking docs the most and were also deeper in the project.
They might look like a bit randomly chosen.

What folder should the focus be on?

mc-dorzo:
[The Parlant Engine has the concepts of GuidelineMatcher, ToolCaller, and MessageComposer.](https://www.parlant.io/docs/engine-internals/overview/)
You can find more detailed documentation there, which should help you understand the architecture better.

A good place to start is with the GuidelineMatcher:
`src/parlant/core/engines/alpha/guideline_matching/guideline_matcher.py`

Then you can move on to the ToolCaller:
`src/parlant/core/engines/alpha/tool_event_generator.py`
`src/parlant/core/engines/alpha/tool_calling/tool_caller.py`

Finally, explore the MessageComposer:
`src/parlant/core/engines/alpha/message_event_composer.py`
`src/parlant/core/engines/alpha/canned_response_generator.py`
`src/parlant/core/engines/alpha/message_generator.py`

It’s worth mentioning again that many parts of the engine are non-trivial. Even if you focus on just one of these components, that would already be a great contribution.

I’m happy to help — feel free to reach out here or on Discord anytime!

kichanyurd:
Closing due to inactivity 😓

PR #549, opened 2025-09-11
fix:deepseek

fucktx:


kichanyurd:
Hi mate @fucktx changes are valid but please undo the formatting changes and keep all the functional ones.

CaptainJi:
@fucktx Modifying the GLMService reference and glm function in sdk.py will result in glm being unavailable.

PR #569, opened 2025-09-23
feat: changeable customer_id & agent_id in session

hochmannA1:
As discussed on [Discord](https://discord.com/channels/1312378700993663007/1415285480500760626), this PR adds the ability to update the `customer_id` and `agent_id` within an active session. This change is particularly helpful in cases where a user logs in on the frontend and their session should reflect the new user. It provides more flexibility for managing session data and ensures that user changes are properly handled during ongoing sessions.


kichanyurd:
@hochmannA1 Great work, thank you!

I'm merging this, but please keep in mind that for now you'd need to use this via a direct HTTP request. In order for this to find its way to our native clients (in Python and TypeScript), we need to update those packages as well, which will happen for the next released version.

But meanwhile, your stopgap solution would be the manual HTTP request.

PR #581, opened 2025-10-04
Relax dependencies

nurikk:
Relax pinned dependencies, allowing adding parlant to existing projects


kichanyurd:
This is a good point, thank you.

Could you please specify which specific dependencies you ran into issues with?

nurikk:
> This is a good point, thank you.
> 
> Could you please specify which specific dependencies you ran into issues with?

There were no particular issues with dependencies, just the issue with pinned versions. For example specific fast api version `fastapi = "0.115.12"`, all can be seen here.

https://github.com/emcie-co/parlant/pull/581/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711


I could use something like `--force` flag installing dependencies, but I'd rather not

kichanyurd:
Hey mate @nurikk could you please revert files changed just due to formatting?

Let's keep the PR focused on only the right files, and move forward with merging it 🙏

nurikk:
> Hey mate @nurikk could you please revert files changed just due to formatting?
> 
> Let's keep the PR focused on only the right files, and move forward with merging it 🙏

done!

kichanyurd:
@nurikk One last request: Can you please remove `poetry.lock` from the PR? It will always tend to create conflicts, yet I can just run `poetry.lock` myself in a separate commit, which is easier.

PR #582, opened 2025-10-05
Feature/mc 967 opentelemetry metrics

mc-dorzo:


src/parlant/core/common.py
```diff
@@ -80,6 +81,18 @@ class DefaultBaseModel(BaseModel):
 ]
 """A JSON-serializable value."""
 
+AttributeValue = Union[
```
kichanyurd: What's this and why's it in common.py?

src/parlant/sdk.py
```diff
@@ -1665,7 +1666,7 @@ class RetrieverContext:
     server: Server
     container: Container
     logger: Logger
-    correlator: ContextualCorrelator
+    tracer: Tracer
```
kichanyurd, line 1692: ***Anywhere*** where we publicly exposed `correlator` (e.g., event emitter, hooks, loaded context, etc.), in order to break people's programs, we need to also expose a property

```python
@property
def correlator(self) -> Tracer:
  self.logger.warning(f"`correlator` is deprecated. Please change your code to use the `tracer` property")
```

That's one.

Second, in the `Tracer` object itself we need to add a property

```python
@property
def correlation_id(self) -> str:
  # Also log something similar here - in whatever way's possible
  return self.trace_id
```

src/parlant/core/tracer.py
```diff
@@ -0,0 +1,169 @@
...
+from contextlib import contextmanager
+import contextvars
+from typing import Iterator, Mapping
+from typing_extensions import override
+
+from parlant.core.common import AttributeValue, generate_id
+
+_UNINITIALIZED = 0xC0FFEE
+
+
```
kichanyurd, line 36: Let's add an alias with a deprecation warning

```python
from typing_extensions import deprecated

@deprecated("Please use the Tracer class instead of ContextualCorrelator")
class ContextualCorrelator(Tracer):
 pass
```

kichanyurd: Also export this alias in `sdk.py` under `__all__`

src/parlant/bin/server.py
```diff
@@ -288,59 +296,69 @@ def load_nlp_service(name: str, extra_name: str, class_name: str, module_path: s
...
-def load_deepseek() -> NLPService:
+def load_deepseek(meter: Meter) -> NLPService:
     return load_nlp_service(
-        "DeepSeek", "deepseek", "DeepSeekService", "parlant.adapters.nlp.deepseek_service"
+        meter, "DeepSeek", "deepseek", "DeepSeekService", "parlant.adapters.nlp.deepseek_service"
     )
 
 
-def load_gemini() -> NLPService:
+def load_gemini(meter: Meter) -> NLPService:
```
kichanyurd: Let's just go ahead and pass the container to these load_* functions, including `load_nlp_service`)

src/parlant/bin/server.py
```diff
@@ -400,6 +418,38 @@ async def load_modules(
                 await shutdown_module()
 
 
+async def _define_logger(container: Container) -> None:
+    if OtelLogger.is_environment_set():
```
kichanyurd: Let's not even load the Otel module if the environment isn't set. Right now we load it because it's a helper function inside it - which drags in all of the Otel dependencies into memory even when not used.

src/parlant/bin/server.py
```diff
@@ -400,6 +418,38 @@ async def load_modules(
...
+        container[Logger] = CompositeLogger([LOGGER, container[WebSocketLogger]])
+
+
+async def _define_tracer(container: Container) -> None:
+    if OtelTracer.is_environment_set():
+        print("OpenTelemetry tracing is enabled.")
+        container[Tracer] = await EXIT_STACK.enter_async_context(OtelTracer())
+
+    else:
+        container[Tracer] = Singleton(LocalTracer)
```
kichanyurd: Again - this is no longer how we define singletons. See define_singleton here

src/parlant/bin/server.py
```diff
@@ -400,6 +418,38 @@ async def load_modules(
...
+        )
+
+    else:
+        container[Logger] = CompositeLogger([LOGGER, container[WebSocketLogger]])
+
+
+async def _define_tracer(container: Container) -> None:
+    if OtelTracer.is_environment_set():
+        print("OpenTelemetry tracing is enabled.")
+        container[Tracer] = await EXIT_STACK.enter_async_context(OtelTracer())
```
kichanyurd: Let's rename the classes from `Otel*` to `OpenTelemetry*` (e.g., `OpenTelemetryTracer`)

src/parlant/core/common.py
mc-dorzo: `AttributeValue` defines the value types supported by OpenTelemetry.
I’ve moved it to `tracer.py` for clarity.

src/parlant/sdk.py
mc-dorzo: Done. I can't add a logger to Tracer since Logger is dependent on Tracer

Issue #586, opened 2025-10-08
[Enhancement] [FR] Allow specifying model_name directly in code and add model selection in UI settings


bedead:
# Motivation

Currently, the model_name for NLP services must be set through environment variables. This can make experimentation or runtime configuration cumbersome — especially when working with multiple providers or testing different model setups programmatically.
I also tried the new chat UI and it’s really sleek — great work! However, it would be even more convenient if the UI had a settings tab that allowed switching between available models directly, rather than relying on environment configuration.

# Solution Proposal

- Backend: Add support for specifying model_name directly in code (e.g., via an argument when initializing the service or through a config object). The env variable could still act as a fallback for backward compatibility.
- Frontend: Add a small “Settings” tab or modal in the chat UI that allows users to:
  - Change the model_name dynamically.
  - Fetch and display available models from NLPServices.<provider_name>.
  - Persist user choice temporarily (e.g., in local storage or session).
This would improve developer flexibility and make the UI more interactive for testing and demos.

# Discussion

Would love to know if this aligns with the maintainers’ roadmap. I’d be happy to help contribute this feature — both on the backend and frontend sides if the proposal seems interesting.

kichanyurd:
Hey @bedead thanks for this initiative!

Let's move forward with this - we'd be excited for you to contribute here.

Can you start by proposing a design + usage example for the backend? Once that looks good, we can discuss the frontend aspect. 🙏

bedead:
Sure, I will start with that, and soon post the design.

bedead:
I’ve started by updating the gemini_service.py provider to support both direct and deferred container injection, so users can call it either as `p.NLPServices.gemini(model_name="gemini-2.0-flash-lite")` or simply `p.NLPServices.gemini` for default fallback to work.

Proposed backend design:
- Each NLP provider (e.g., Gemini, OpenAI, Anthropic) already has a static method in NLPServices that currently accepts Container, I extended it to take in Optional `model_name`, code snippet mentioned below.
- If a `container` is passed, it directly returns the configured NLPService.
- If `model_name` is passed, it loads the GeminiService with desired model_name.
- If nothing is passed, it works just as before.

This keeps the API consistent and declarative while ensuring backward compatibility.


```
# in sdk.py

    @staticmethod
    def gemini(container: Optional[Container] = None, model_name: Optional[str] = None) -> NLPService:
        """
        Returns a callable that creates a Gemini NLPService instance using the provided container and model_name.
[...]

Issue #584, opened 2025-10-07
[Bug]  How to set a particular model using Gemini api


amir-shehzad:
I haven't been able to find in the documentation and in the code `src/parlant/adapters/nlp/gemini_service.py`
and `src/parlant/sdk.py`
how to set a particular model like gemini-2.5-flash

mc-dorzo:
Hey @amir-shehzad 👋

If you specifically want to use Gemini-2.5-flash when running the Parlant server, that’s actually the default model used when Parlant runs with Gemini as the NLP Service.


<img width="769" height="173" alt="Image" src="https://github.com/user-attachments/assets/3135e19b-ed8c-431d-85c6-a3ade08f8e8a" />

bedead:
From the docs at https://www.parlant.io/docs/quickstart/installation/.
But this only specify the model provider not a specific model_name.

> Using Other LLM Providers
> Parlant supports multiple LLM providers by default, accessible via the p.NLPServices class. You can also add your own provider by implementing the p.NLPService interface, which you can learn how to do in the  section.
> 
> To use one of the built-in-providers, you can specify it when creating the server. For example:
> 
> async with p.Server(nlp_service=p.NLPServices.cerebras) as server:
>   ...
> 
> Note that you may need to install an additional "extra" package for some providers. For example, to use the Cerebras NLP service:
> 
> pip install "parlant[cerebras]"
> 
> Having said that, Parlant is observed to work best with  and  models, as these models are highly consistent in generating high-quality completions with valid JSON schemas—so we recommend using one of those if you're just starting out.


One way I just found according to docs is to use vertexai provider and to choose a specific model define it in env variables as below:
```
VERTEX_AI_PROJECT_ID=your-gcp-project-id
VERTEX_AI_REGION=us-central1  # Put your region
VERTEX_AI_MODEL=claude-opus-4 # or gemini-2.5-flash
```

But this way doesn't seem very ideal to me.

ghost:
@mc-dorzo Thanks for the quick response. I was asking generally how you do provide model names to use in parlant?
like if i wanted to use gemini-2.0-flash or gemini-2.5-pro or other models in gemini, how do yo do that?

bedead:
Actually, I don't think it's yet possible to specific it from direct code. I have just created an [FR] issue #586 about this, ones it gets accepted, I will start working on it.

[...]

PR #591, opened 2025-10-10
feat(gemini): add specific multi-model selection and fallback support issue #586

bedead:
## Description
#586 #584 This PR extends the GeminiService provider to support multiple model names with automatic fallback handling.

## Key Updates

Added support for passing model_name as:

- a single string (e.g., "gemini-2.0-flash-lite")
- a list of model names (e.g., ["gemini-2.0-flash-lite", "gemini-2.5-flash"])
- or left None, for default model fallback behavior.

Integrated FallbackSchematicGenerator to automatically switch to the next model in the list if one fails or is overloaded.

Kept backward compatibility — existing usages like p.NLPServices.gemini() or None continue to work.

## Example Usage
```
import asyncio
import parlant.sdk as p

async def main():
    async with p.Server(
        nlp_service=p.NLPServices.gemini(
            model_name=["gemini-2.0-flash-lite", "gemini-2.5-flash"]
        )
    ) as server:
        agent = await server.create_agent(
            name="Otto Carmen",
            description="You work at a car dealership",
        )

asyncio.run(main())

```

## Files Modified

```
parlant/adapters/nlp/gemini_service.py
parlant/sdk.py
```

## Files Added 

`examples/main.py`


examples/main.py
```diff
@@ -0,0 +1,18 @@
+import asyncio
```
mc-dorzo: Please delete this file (I guess it was for testing)

src/parlant/adapters/nlp/gemini_service.py
```diff
@@ -12,16 +12,19 @@
...
 import os
 import time
 import types
+import types
 from google.api_core.exceptions import NotFound, TooManyRequests, ResourceExhausted, ServerError
 import google.genai  # type: ignore
 import google.genai.types  # type: ignore
 from collections.abc import Mapping as MappingABC, Sequence as SequenceABC
-from typing import Any, Literal, Mapping, Sequence, Union, cast
+from typing import Any, List, Literal, Mapping, Optional, Sequence, Union, cast
```
mc-dorzo: The list here is not necessary. Please replace with regular `list`all uses

src/parlant/adapters/nlp/gemini_service.py
```diff
@@ -136,12 +139,26 @@ async def generate(
...
+            **gemini_api_arguments,  # type: ignore
+        )
+
         t_start = time.time()
         try:
             response = await self._client.aio.models.generate_content(
                 model=self.model_name,
                 contents=prompt,
                 config=config,
+                config=config,
```
mc-dorzo: duplication of the line above it

src/parlant/adapters/nlp/gemini_service.py
```diff
@@ -164,6 +190,7 @@ async def generate(
 
         try:
             model_content = self.schema.model_validate(json_result)
+            model_content = self.schema.model_validate(json_result)
```
mc-dorzo: duplication of line above it

src/parlant/adapters/nlp/gemini_service.py
```diff
@@ -406,21 +462,59 @@ def verify_environment() -> str | None:
...
     @override
     async def get_schematic_generator(self, t: type[T]) -> GeminiSchematicGenerator[T]:
-        return FallbackSchematicGenerator[t](  # type: ignore
-            Gemini_2_5_Flash[t](self._logger),  # type: ignore
-            Gemini_2_5_Pro[t](self._logger),  # type: ignore
+        selected_model = self._model_name
+
+        if isinstance(selected_model, list):
+            model_classes: List[SchematicGenerator][T] = []
+            for model_name in selected_model:
```
mc-dorzo: Is it a singular model or multiple? Please fix the naming

src/parlant/sdk.py
```diff
@@ -291,14 +294,23 @@ def together(container: Container) -> NLPService:
         return TogetherService(container[Logger])
 
     @staticmethod
-    def gemini(container: Container) -> NLPService:
-        """Creates a Gemini NLPService instance using the provided container."""
+    def gemini(container: Optional[Container] = None, model_name: Optional[Union[List[str], str]] = None) -> NLPService:
+        """
+        Returns a callable that creates a Gemini NLPService instance using the provided container and model_name.
```
mc-dorzo: Since the documentation here requires more maintenance in case we change the model, let's keep it as simple as it was before

src/parlant/sdk.py
```diff
@@ -39,6 +39,8 @@
 from rich.text import Text
 from types import TracebackType
 from typing import (
+    List,
```
mc-dorzo: Again, the List is not necessary , change it to regular `list`

src/parlant/sdk.py
```diff
@@ -291,14 +294,23 @@ def together(container: Container) -> NLPService:
         return TogetherService(container[Logger])
 
     @staticmethod
-    def gemini(container: Container) -> NLPService:
-        """Creates a Gemini NLPService instance using the provided container."""
+    def gemini(container: Optional[Container] = None, model_name: Optional[Union[List[str], str]] = None) -> NLPService:
```
mc-dorzo: Optional need to be replaces with `Container | None` and `Union[list[str], str] | None`

mc-dorzo: Then delete the Optional import

src/parlant/sdk.py
```diff
@@ -291,14 +294,23 @@ def together(container: Container) -> NLPService:
...
         from parlant.adapters.nlp.gemini_service import GeminiService
 
-        if error := GeminiService.verify_environment():
-            raise SDKError(error)
+        def factory(c: Container) -> NLPService:
+            if error := GeminiService.verify_environment():
+                raise SDKError(error)
+            return GeminiService(c[Logger], model_name=model_name)
+
+        if container is not None:
```
mc-dorzo: Not sure I understand he logic here. In what case is the container None?

bedead: The container is None when the service is being registered or configured—before the server or dependency injector has created the full container. In that case, the method returns a callable (factory) that will receive the container later to actually create the service.

bedead:
Should I update the docs to?

bedead:
Here's screenshots as asked, thanks for comments.
## Test code:
image


## ChatUI:
image


## Terminal:
image

mc-dorzo:
Great job @bedead !
Can you please add a doc describing usage at `https://github.com/emcie-co/parlant/tree/develop/docs/adapters/nlp/gemini.md`

bedead:
@mc-dorzo I was also thinking about adding this feature to other nlp adapters. Should I add them as well, and create separate PR for each?

kichanyurd:
@bedead please also see my comment in the issue you opened.

Regarding other providers... that would be truly lovely :) And yes, separate PRs please!

Great work mate - we appreciate your initiative a lot!

bedead:
@kichanyurd Great that you are well now (I know, sickness drains a lot). Also, I just viewed your comment in previous issue. I will make the change.

kichanyurd:
@bedead LGTM! Can you confirm this is well-tested and ready to go?

src/parlant/sdk.py
```diff
@@ -292,15 +293,21 @@ def together(container: Container) -> NLPService:
         return TogetherService(container[Logger])
 
     @staticmethod
-    def gemini(container: Container) -> NLPService:
-        """Creates a Gemini NLPService instance using the provided container."""
+    def gemini(container: Container | None = None, generative_model_name: Union[list[str], str] | None = None) -> NLPService:
```
kichanyurd: Isn't there a lint issue there, as we don't always return a proper NLPService, but rather sometimes we return a factory method?

bedead: My bad, didn't run linting. Fixed it with

`    def gemini(container: Container | None = None, generative_model_name: Union[list[str], str] | None = None) -> Union[NLPService, Callable[[Container], NLPService]]:`

bedead:
@kichanyurd  SS:
image
image

Works without passing in `generative_model_name` params also.
image
image

kichanyurd:
@bedead A realization came to me with a bit of a delay 😅 

I think this pattern will be useful to virtually *all* NLP services.

So we need to ensure this is the right way we want to do it. It needs to be flexible enough so that we don't end up changing the pattern in a few weeks/months (consequently breaking the API for many users).

Please allow me a few more days to think it over! Appreciate your patience. 🙏

bedead:
@kichanyurd I understand. It's important to think about this properly. I was also thinking, that it would be more intuitive and reasonable to set models specific to agents rather than whole server.

Concept wise, this will allow users to choose from multiple providers (which could be specified in server). And then select specific models in agents. As certain providers and models are good with certain tasks. E.g. From my experience gpt models performs well in debugging and generating code templates. Whereas gemini models are great with image manipulation, and bit of creative work. And Antrophic models are good at understand code, generating new code.

Current implementation will make provider as will as models very static. 

One way that is used by most of agentic frameworks, is that llm could be initialised (with providers and models name) and then passed to specific agent.

kichanyurd:
@bedead, sorry for the late response here. We're totally swamped right now with work on v3.1 and trying to get SLMs under control to bring costs down as close to the floor as we can.

I like your idea of controlling `NLPService` per agent. I added it to our roadmap and will try to get it into v3.1 if I can.

I'll also leave this issue open for now until we can come up with an idea to configure each `NLPService` in an elegant and uniform manner. I was thinking perhaps the [Builder pattern](https://en.wikipedia.org/wiki/Builder_pattern) might be a good direction to explore here.

bedead:
@kichanyurd no issues, if there are any specific areas which needs help. You can ping me; I would love to contribute more to this project.

PR #602, opened 2025-10-19
Feature/mc 1417 small disambiguation prompt edits

MCBarKar:
Clarified the task description section of the disambiguation prompt.
The edits are only rephrasings of the old rules - no change in behavior is expected.

Also added one test based on a case where a failure was previously reported (on my end the test passes both before and after this PR, but kept it in).

@HadarYosef1 Could you let me know if the edits look ok? The tests all pass.
Please wait for Hadar's approval before merging.

### DIFF
--- a/src/parlant/core/engines/alpha/guideline_matching/generic/disambiguation_batch.py
+++ b/src/parlant/core/engines/alpha/guideline_matching/generic/disambiguation_batch.py
@@ -150,10 +150,6 @@ class GenericDisambiguationGuidelineMatchingBatch(GuidelineMatchingBatch):
                         prompt=prompt,
                         hints={"temperature": generation_attempt_temperatures[generation_attempt]},
                     )
-                    with open("disambiguation prompt.txt", "w") as f:
-                        f.write(prompt.build())
-                    with open("disambiguation inference.txt", "w") as f:
-                        f.write(inference.content.model_dump_json(indent=2))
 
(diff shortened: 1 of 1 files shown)


src/parlant/core/engines/alpha/guideline_matching/generic/disambiguation_batch.py
```diff
@@ -296,34 +297,38 @@ def _build_prompt(
 
 Task Description
 ----------------
-Sometimes a customer expresses that they’ve experienced something or want to proceed with something, but there are multiple possible ways to go, and it’s as-yet unclear what exactly they intend.
-In such cases, we need to identify the potential options and ask the customer which one they mean.
+During your interaction with the customer, they may express a need or problem that could potentially be handled by multiple guidelines, creating ambiguity.
```
HadarYosef1, line 300: We usually define what a guideline is at the beginning of the task description. I think I tried to describe the task here without defining it so it may got a little too vague. Do you think its clear from the context here what do you mean by guideline?

MCBarKar: Yes - there's a GENERAL DESCRIPTION section just before the TASK DESCRIPTION that explains what a a guideline is. It doesn't appear in the changes log because I didn't touch it.

MCBarKar:
Can be merged

PR #604, opened 2025-10-20
Refactor Gemini service to update embedding model names and configurations

mrgehlot:
- Changed model name from "text-embedding-004" to "gemini-embedding-001" in GoogleEstimatingTokenizer and GeminiTextEmbedding_001 classes.
- Adjusted max_tokens from 8000 to 2048 and dimensions from 768 to 3072 in GeminiTextEmbedding_001.
- Updated get_embedder method in GeminiService to return the new GeminiTextEmbedding_001 instance.

Signed-off-by: Author Name <abhigelot123@gmail.com>

### DIFF
--- a/src/parlant/adapters/nlp/gemini_service.py
+++ b/src/parlant/adapters/nlp/gemini_service.py
@@ -64,7 +64,7 @@ class GoogleEstimatingTokenizer(EstimatingTokenizer):
     @override
     async def estimate_token_count(self, prompt: str) -> int:
         model_approximation = {
-            "text-embedding-004": "gemini-2.5-flash",
+            "gemini-embedding-001": "gemini-2.5-flash",
         }.get(self._model_name, self._model_name)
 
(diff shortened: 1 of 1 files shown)


kichanyurd:
@mrgehlot Nice one! Can you confirm this is currently tested and working as expected?

Out of curiosity, the previous implementation didn't work at all? Or did you just go for a different version of the embedder?

mrgehlot:
@kichanyurd  Yes, it's tested and working well!
The old `text-embedding-004` was working fine, but I switched to `gemini-embedding-001` because:
Better performance - it scores higher on benchmarks
More languages - supports 100+ languages vs the old model's limited support
Longer text - can handle up to 2,048 tokens
Future-proof - Google is deprecating `text-embedding-004` in `January 2026`
So it wasn't broken, just wanted to upgrade to the newer, better model before the old one gets discontinued.

PR #638, opened 2025-11-14
Integrate Qdrant as a Vector Database

Pro-Sifat-Hasan:
## Summary
This PR introduces Qdrant integration as a vector database to enhance vector search and retrieval capabilities.

## Details
- Added Qdrant as a supported vector database.
- Implemented initialization, query, and search operations for Qdrant.
- Updated configuration logic to allow switching between existing retrievers and Qdrant.
- Provided examples and documentation for Qdrant setup and usage.
- Ensured seamless integration with existing retrieval workflows.

## Type of Change
- [x] ✨ Enhancement (new feature or improvement)
- [x] 🐞 Bug fix
- [ ] 🧩 Refactor
- [x] 📚 Documentation update

## Testing
- Verified Qdrant connection and query execution.
- Tested vector retrieval workflows end-to-end.
- Confirmed fallback logic for non-Qdrant retrievers.

## Additional Notes
This integration enables developers to leverage Qdrant for fast and scalable vector search, improving overall retrieval efficiency and flexibility.

### DIFF
--- a/docs/adapters/vector_db/qdrant.md
+++ b/docs/adapters/vector_db/qdrant.md
@@ -13,10 +13,6 @@ For general Parlant usage, see the [official documentation](https://www.parlant.
 
 ### Setup (Manual)
 
-Set up Qdrant and vector stores using `AsyncExitStack`:
-
-#### Local Storage
-
 ```python
 import parlant.sdk as p
 from pathlib import Path
@@ -46,6 +42,17 @@ async def configure_container(container: p.Container) -> p.Container:
             embedding_cache_provider=lambda: container[EmbeddingCache],
         )
     )
+    
+    # For Qdrant Cloud, replace the above with:
+    # qdrant_db = await exit_stack.enter_async_context(
+    #     QdrantDatabase(
+    #         logger=container[Logger],
+    #         url="https://your-cluster-id.us-east4-0.gcp.cloud.qdrant.io",
+    #         api_key="your-api-key-here",
+    #         embedder_factory=EmbedderFactory(container),
+    #         embedding_cache_provider=lambda: container[EmbeddingCache],
+    #     )
+    # )
     
     # Configure stores using vector database
     container[GlossaryStore] = await exit_stack.enter_async_context(
@@ -105,99 +112,6 @@ async def main():
         print(f"Created term: {term.name}")
         # All vector operations now use Qdrant
 ```
-
-#### Qdrant Cloud
-
-```python
-import parlant.sdk as p
-from contextlib import AsyncExitStack
-from parlant.adapters.vector_db.qdrant import QdrantDatabase
-from parlant.core.nlp.embedding import EmbedderFactory, EmbeddingCache, Embedder
-from parlant.core.loggers import Logger
-from parlant.core.nlp.service import NLPService
-from parlant.core.glossary import GlossaryVectorStore, GlossaryStore
-from parlant.core.canned_responses import CannedResponseVectorStore, CannedResponseStore
-from parlant.core.capabilities import CapabilityVectorStore, CapabilityStore
-from parlant.core.journeys import JourneyVectorStore, JourneyStore
-from parlant.adapters.db.transient import TransientDocumentDatabase
-
-async def configure_container(container: p.Container) -> p.Container:
-    embedder_factory = EmbedderFactory(container)
-
-    async def get_embedder_type() -> type[Embedder]:
-        return type(await container[NLPService].get_embedder())
-    
-    exit_stack = AsyncExitStack()
-    qdrant_db = await exit_stack.enter_async_context(
-        QdrantDatabase(
-            logger=container[Logger],
-            url="https://your-cluster-id.us-east4-0.gcp.cloud.qdrant.io",
-            api_key="your-api-key-here",
-            embedder_factory=EmbedderFactory(container),
-            embedding_cache_provider=lambda: container[EmbeddingCache],
-        )
-    )
-    
-    # Configure stores using vector database
-    container[GlossaryStore] = await exit_stack.enter_async_context(
-        GlossaryVectorStore(
-            id_generator=container[p.IdGenerator],
-            vector_db=qdrant_db,
-            document_db=TransientDocumentDatabase(),
-            embedder_factory=embedder_factory,
-            embedder_type_provider=get_embedder_type,
-        )  # type: ignore
-    )
-    
-    container[CannedResponseStore] = await exit_stack.enter_async_context(
-        CannedResponseVectorStore(
-            id_generator=container[p.IdGenerator],
-            vector_db=qdrant_db,
-            document_db=TransientDocumentDatabase(),
-            embedder_factory=embedder_factory,
-            embedder_type_provider=get_embedder_type,
-        )  # type: ignore
-    )
-    
-    container[CapabilityStore] = await exit_stack.enter_async_context(
-        CapabilityVectorStore(
-            id_generator=container[p.IdGenerator],
-            vector_db=qdrant_db,
-            document_db=TransientDocumentDatabase(),
-            embedder_factory=embedder_factory,
-            embedder_type_provider=get_embedder_type,
-        )  # type: ignore
-    )
-    
-    container[JourneyStore] = await exit_stack.enter_async_context(
-        JourneyVectorStore(
-            id_generator=container[p.IdGenerator],
-            vector_db=qdrant_db,
-            document_db=TransientDocumentDatabase(),
-            embedder_factory=embedder_factory,
-            embedder_type_provider=get_embedder_type,
-        )  # type: ignore
-    )
-    
-    return container
-
-async def main():
-    async with p.Server(configure_container=configure_container) as server:
-        agent = await server.create_agent(
-            name="My Agent",
-            description="Agent using Qdrant Cloud for persistent storage",
-        )
-        
-        # Test: Create a term to verify Qdrant is working
-        term = await agent.create_term(
-            name="Example Term",
-            description="This is stored in Qdrant Cloud",
-        )
-        print(f"Created term: {term.name}")
-        # All vector operations now use Qdrant Cloud
-```
-
-**Note:** The `AsyncExitStack` ensures proper cleanup of async context managers. The Server automatically manages cleanup when it exits. The `document_db=TransientDocumentDatabase()` is correct for vector stores—it handles non-vector metadata (like tag associations), while vector data goes to Qdrant.
 
 
 ## Verification
--- a/src/parlant/adapters/vector_db/qdrant.py
+++ b/src/parlant/adapters/vector_db/qdrant.py
@@ -108,60 +108,6 @@ async def _retry_on_timeout_async(
     raise RuntimeError("Retry logic failed unexpectedly")
 
 
-async def _retry_on_timeout_sync(
-    operation: Callable[[], T],
-    max_retries: int = 3,
-    logger: Optional[Logger] = None,
-) -> T:
-    """
-    Retry a synchronous operation on timeout errors with exponential backoff.
-    This is an async function that wraps sync operations for retry logic.
-
-    Args:
-        operation: The synchronous operation to retry (callable that returns T)
-        max_retries: Maximum number of retry attempts
-        logger: Optional logger for warning messages
-
-    Returns:
-        The result of the operation
-
-    Raises:
-        The last exception if all retries fail
-    """
-    last_exception: Exception | None = None
-
-    for attempt in range(max_retries):
-        try:
-            return operation()
-        except (ResponseHandlingException, Exception) as e:
-            # Check if it's a timeout error
-            error_str = str(e).lower()
-            is_timeout = (
-                "timeout" in error_str
-                or "read operation timed out" in error_str
-                or "readtimeout" in error_str
-            )
-
-            if is_timeout and attempt < max_retries - 1:
-                delay = 1.0 * (2**attempt)  # Exponential backoff: 1s, 2s, 4s
-                if logger:
-                    logger.warning(
-                        f"Qdrant operation timed out (attempt {attempt + 1}/{max_retries}). "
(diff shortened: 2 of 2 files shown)


docs/adapters/vector_db/qdrant.md
```diff
@@ -0,0 +1,752 @@
...
+
+### Data Persistence
+
+- **In-memory**: Data lost on process exit
+- **File system**: Data persists in specified directory
+- **Remote server**: Data persists on Qdrant server
+
+---
+
+## 📊 Comparison with Other Adapters
```
mc-dorzo: For maintenance purposes, let's remove the sections that are common to every adapter. We will need to update those sections (and remember to do so). Let's leave the decision of choosing the technology to the developers.

pyproject.toml
```diff
@@ -80,6 +81,7 @@ litellm = { version = "^1.61.16", optional = true }
 pymongo = { version = "^4.11.1", optional = true }
 fireworks-ai = { version = "^0.19.19", optional = true }
 mistralai = { version = "^1.0.0", optional = true }
+pydantic = "^2.12.4"
```
mc-dorzo: Do we need to add this?

src/parlant/adapters/vector_db/qdrant.py
```diff
@@ -0,0 +1,1241 @@
...
+    return hash_value % (2**63 - 1)
+
+
+def _convert_where_to_qdrant_filter(where: Where) -> Optional[Filter]:
+    """Convert a Where filter to a Qdrant Filter."""
+    if not where:
+        return None
+
+    # Handle logical operators
+    if "$and" in where:
```
mc-dorzo, line 157: Can you add tests for different logical operators and combinations between the two?

src/parlant/adapters/vector_db/qdrant.py
```diff
@@ -0,0 +1,1241 @@
...
+                        embeddings = list(
+                            (await embedder.embed([cast(str, unembedded_doc["content"])])).vectors
+                        )
+
+                    self.qdrant_client.upsert(
+                        collection_name=embedded_collection_name,
+                        points=[
+                            models.PointStruct(
+                                id=embedded_point.id,  # Keep existing point ID
+                                vector=embeddings[0],
```
mc-dorzo: There's a bug here - embeddings at this point may not have been assigned in case the condition in line 444 was not met.

src/parlant/adapters/vector_db/qdrant.py
```diff
@@ -0,0 +1,1241 @@
...
+                ids=[version_point_id],
+                with_payload=True,
+            )
+            if points and points[0].payload:
+                return cast(int, points[0].payload.get("version", 1))
+        except Exception:
+            pass
+        return 1
+
+    async def _set_collection_version(self, collection_name: str, version: int) -> None:
```
mc-dorzo, line 439: Currently, if you run the find function, you will retrieve this metadata, which is not relevant, and the collection will become inconsistent. I'd suggest using the existing collection of metadata for this.

docs/adapters/vector_db/qdrant.md
Pro-Sifat-Hasan: Deleted the qdrant.md file as suggested.

pyproject.toml
Pro-Sifat-Hasan: We don’t need this, so I’ve removed it.

src/parlant/adapters/vector_db/qdrant.py
Pro-Sifat-Hasan: Added the tests for logical operators and their combinations.

src/parlant/adapters/vector_db/qdrant.py
Pro-Sifat-Hasan: Good catch — I’ve fixed the bug.
Now, if the checksum has changed, new embeddings are generated; otherwise, it uses the existing vector from embedded_point.vector.

src/parlant/adapters/vector_db/qdrant.py
Pro-Sifat-Hasan: Solved — switched to using the existing metadata collection so it no longer appears in find results.

Pro-Sifat-Hasan:
@mc-dorzo  Thanks for the review! I’ve fixed all the issues, added improvements, and updated the changes. Please take a look again, and if everything is good, feel free to merge.

docs/adapters/vector_db/qdrant.md
```diff
@@ -0,0 +1,487 @@
...
+### Collection Naming
+
+- Unembedded: `{name}_unembedded`
+- Embedded: `{name}_{EmbedderType.__name__}` (e.g., `documents_OpenAITextEmbedding3Large`)
+
+**Note**: Collections are automatically created and synchronized. You don't manage them directly.
+
+---
+
+## Usage Patterns
```
mc-dorzo: Since it already implements the VectorDatabase interface, I wouldn’t include the guidance below—there’s nothing Qdrant-specific to add. As a general rule, this documentation should focus primarily on integrating Qdrant with Parlant, rather than explaining how to use Parlant itself or what Qdrant is.

docs/adapters/vector_db/qdrant.md
```diff
@@ -0,0 +1,487 @@
...
+                "topic": "hydration",
+            },
+        ]
+        
+        for doc in medical_docs:
+            await collection.insert_one(doc)
+
+
+@p.tool
+async def search_medical_knowledge(
```
mc-dorzo: It is not a good practice to create Qdrant client and collection every  tool execution.
I wouldn't put such use case. 
I would delete the whole section of usage, and instead just add how to actually use the Qdrant db in parlant:
You should define [`configure_container`](https://www.parlant.io/docs/advanced/engine-extensions/#registering-components) and then implement a custom store that needs the vector stores. Please validate it works for you. (:

tests/adapters/vector_db/test_qdrant.py
```diff
@@ -0,0 +1,1259 @@
...
+        new_collection = await qdrant_database.get_or_create_collection(
+            "test_collection",
+            _TestDocumentV2,
+            embedder_type=OpenAITextEmbedding3Large,
+            document_loader=_document_loader,
+        )
+
+        new_documents = await new_collection.find({})
+        assert len(new_documents) == 3
+        assert new_documents[0]["id"] == ObjectId("1")
```
mc-dorzo: Are we sure that this query in Qdrant will return in the order it was created? (I don't think so. try to run this test few times)

src/parlant/adapters/vector_db/qdrant.py
```diff
@@ -0,0 +1,1210 @@
...
+                try:
+                    if loaded_doc := await document_loader(prospective_doc):
+                        if loaded_doc != prospective_doc:
+                            # Update the unembedded collection
+                            self.qdrant_client.upsert(
+                                collection_name=unembedded_collection_name,
+                                points=[
+                                    models.PointStruct(
+                                        id=point.id,
+                                        vector=[],
```
mc-dorzo: This is a bug, you can transfer `vector=[]`, it should be like in the `insert_one` (`vector=[0]`)

src/parlant/adapters/vector_db/qdrant.py
```diff
@@ -0,0 +1,1210 @@
...
+                            ),
+                        )
+                        failed_migrations.append(prospective_doc)
+
+                except Exception as e:
+                    self._logger.error(f"Failed to load document '{prospective_doc}'. error: {e}.")
+                    failed_migrations.append(prospective_doc)
+
+            # Store failed migrations in a separate collection for debugging
+            if failed_migrations:
```
mc-dorzo, line 406: You started here with using (correctly IMO) our DocumentCollection, then you upsert directly from Qdrant. it will lead to unexpected behaviors

mc-dorzo:
Hey dude, just wrote you few more comments (: 

good job on fixing so quickly btw, as always :)

docs/adapters/vector_db/qdrant.md
Pro-Sifat-Hasan: Agreed. I’ve removed the generic usage patterns and refocused the documentation strictly on the Parlant configuration and integration steps.

docs/adapters/vector_db/qdrant.md
Pro-Sifat-Hasan: Good catch on the client instantiation overhead. I’ve completely rewritten this section to demonstrate the correct pattern: defining configure_container and registering the custom store properly.

tests/adapters/vector_db/test_qdrant.py
Pro-Sifat-Hasan: You are right—the default return order isn't guaranteed. I updated the test to explicitly sort the results by ID before asserting, which ensures the test is deterministic.

src/parlant/adapters/vector_db/qdrant.py
Pro-Sifat-Hasan: Fixed. I aligned this with insert_one and set it to vector=[0] to prevent the error.

src/parlant/adapters/vector_db/qdrant.py
Pro-Sifat-Hasan: Valid point. I refactored this block to rely strictly on the DocumentCollection interface rather than bypassing it with a direct Qdrant upsert. This should prevent the unexpected behavior.

Pro-Sifat-Hasan:
Thanks for the quick turnaround and the kind words, @mc-dorzo! I appreciate the detailed eye on the usage patterns and the migration logic. I’ve addressed all the points below, cleaned up the documentation to focus on integration, and fixed the vector bug.

Ready for another look when you are!

docs/adapters/vector_db/qdrant.md
```diff
@@ -0,0 +1,223 @@
...
+```python
+import parlant.sdk as p
+from pathlib import Path
+from parlant.adapters.vector_db.qdrant import QdrantDatabase, setup_qdrant_for_parlant
+from parlant.core.nlp.embedding import EmbedderFactory, EmbeddingCache
+from parlant.core.loggers import Logger
+
+async def configure_container(container: p.Container) -> p.Container:
+    # Create custom QdrantDatabase instance
+    qdrant_db = await QdrantDatabase(
```
mc-dorzo: you configured the qdrant_db, but didnt use it..
do like so
```python
                embedder_factory = EmbedderFactory(c())

                async def get_embedder_type() -> type[Embedder]:
                   return type(await c()[NLPService].get_embedder())
                
                exit_stack = AsyncExitStack()
                qdrant_db = await exit_stack.enter_async_context(
                    QdrantDatabase(
                        logger=container[Logger],
                        path=Path("./qdrant_data"),  # or url/api_key for cloud
                        embedder_factory=EmbedderFactory(container),
                        embedding_cache_provider=lambda: container[EmbeddingCache],
                    )
                )
                
                
                than you can start configuring the stores using vector database such as 
                container[GlossaryStore] = await exit_stack.enter_async_context(
                    GlossaryVectorStore(
                        id_generator=container[IdGenerator],
                        vector_db=qdrant_db,
                        document_db=TransientDocumentDatabase(),
                        embedder_factory=embedder_factory,
                        embedder_type_provider=get_embedder_type,
                    )  # type: ignore
                )
                other stores ...
```

mc-dorzo: and validate it works :) And then close the server and run again! Maybe the document db must not be transient.. test it.

docs/adapters/vector_db/qdrant.md
```diff
@@ -0,0 +1,223 @@
...
+        # Add any custom parameters here
+    ).__aenter__()
+    
+    # Setup all vector stores to use this Qdrant instance
+    return await setup_qdrant_for_parlant(container, qdrant_db)
+```
+
+## How It Works
+
+Qdrant uses two collections per data type:
```
mc-dorzo: This is very low-level implementation details which I'm not sure if it will interest someone

docs/adapters/vector_db/qdrant.md
```diff
@@ -0,0 +1,223 @@
...
+term = await agent.create_term(
+    name="Test Term",
+    description="This should be stored in Qdrant",
+)
+# Then chat with agent about "test term" - it should understand via vector search
+```
+
+---
+
+## Custom Stores
```
mc-dorzo: Not sure why to create a new store.. Please elaborate what you meant in this section

docs/adapters/vector_db/qdrant.md
Pro-Sifat-Hasan: I’ve updated the configuration to use AsyncExitStack as suggested.

docs/adapters/vector_db/qdrant.md
Pro-Sifat-Hasan: Agreed. I’ve removed this section to keep the documentation focused on usage rather than internal implementation details.

docs/adapters/vector_db/qdrant.md
Pro-Sifat-Hasan: Good point. I realized this section was unnecessary for the general setup, so I’ve removed it to avoid confusion.

Pro-Sifat-Hasan:
All updates are completed — removed the redundant Qdrant methods, refreshed the docs, and everything is working after testing. Please review when you get a moment, and if anything else looks good to you, feel free to merge.

src/parlant/adapters/vector_db/qdrant.py
```diff
@@ -0,0 +1,1429 @@
...
+    VectorDatabase,
+    TDocument,
+    identity_loader,
+)
+
+
+T = TypeVar("T")
+
+
+async def _retry_on_timeout_async(
```
mc-dorzo, line 56: Where do you use this? I think you should delete the sync one and use this?

Pro-Sifat-Hasan: Already done — the sync version is removed, async version kept

PR #653, opened 2025-11-26
Custom OpenAI semantic generator for OpenAI-compatible API

huydt-avatarin:
I just want to make parlant work with my self-host vLLM

### DIFF
--- a/src/parlant/adapters/nlp/openai_service.py
+++ b/src/parlant/adapters/nlp/openai_service.py
@@ -112,7 +112,9 @@ class OpenAISchematicGenerator(BaseSchematicGenerator[T]):
 
         self._client = AsyncClient(api_key=os.environ["OPENAI_API_KEY"])
 
-        self._tokenizer = OpenAIEstimatingTokenizer(model_name=tokenizer_model_name or self.model_name)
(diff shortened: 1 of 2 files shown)


src/parlant/adapters/nlp/openai_service.py
```diff
@@ -389,6 +389,125 @@ def max_tokens(self) -> int:
...
+        self,
+        model_name: str,
+        logger: Logger,
+        meter: Meter | None = None,
+        base_url: str | None = None,
+        api_key: str | None = None,
+        tokenizer_model_name: str | None = None,
+        max_tokens: int | None = None,
+    ) -> None:
+        super().__init__(logger=logger, meter=meter or NullMeter(), model_name=model_name)
```
mc-dorzo: You can avoid this or `NullMeter()`

huydt-avatarin: Yeah I removed it

mc-dorzo:
Hey @huydt-avatarin  — we were wondering how you’re actually utilizing it, since it’s not implemented in the OpenAIService.
Also, I noticed quite a bit of duplication in the generate function. Could you possibly leverage and inherit from the OpenAISchematicGenerator class?

PR #656, opened 2025-11-30
changed relative action arq from str to seq[str]

MCBarKar:
Changed the 'conditions' ARQ to be a sequence of strings instead of a string.

The previous type confused smaller language models occasionally. They would output a list and not a string of a list.

### DIFF
--- a/src/parlant/core/services/indexing/relative_action_proposer.py
+++ b/src/parlant/core/services/indexing/relative_action_proposer.py
@@ -32,7 +32,7 @@ class RelativeActionProposition(DefaultBaseModel):
 
 class RelativeActionBatch(DefaultBaseModel):
     index: str
-    conditions: Sequence[str]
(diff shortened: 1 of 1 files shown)


src/parlant/core/services/indexing/relative_action_proposer.py
```diff
@@ -32,7 +32,7 @@ class RelativeActionProposition(DefaultBaseModel):
 
 class RelativeActionBatch(DefaultBaseModel):
     index: str
-    conditions: str
+    conditions: Sequence[str]
```
kichanyurd: For extra robustness, let's define this as `conditions: Sequence[str] | None = None`

This will help with cases where some models decide to place `null` instead of an empty list - I've seen it before.

MCBarKar: Done

PR #665, opened 2025-12-07
S3 adapter

guyna25:
Added make_s3_db into SDK and matching param for DB creation.
Added S3 DB with connection string


pyproject.toml
```diff
@@ -87,6 +87,7 @@ snowflake-connector-python = { version = "^3.12.0", optional = true }
 
 [tool.poetry.group.dev.dependencies]
 ipython = "^8.26.0"
+moto = {extras = ["s3"], version = "^5.0.0"}
```
mc-dorzo: This should be optional, with making the whole feature as extras, please see other NLP services and databases such `mongo` as reference

src/parlant/adapters/db/s3.py
```diff
@@ -0,0 +1,456 @@
...
+        schema: type[TDocument],
+        document_loader: Callable[[BaseDocument], Awaitable[Optional[TDocument]]],
+    ) -> S3DocumentCollection[TDocument]:
+        # Check if collection "exists" by checking if any objects exist with prefix
+        exists = await self._collection_exists(name)
+        if not exists and name not in self._collections:
+            raise ValueError(f'Collection "{name}" does not exists')
+
+        if name not in self._collections:
+            self._collections[name] = S3DocumentCollection(
```
mc-dorzo, line 123: I know we have a Snowflake reference implementation where the migration runs at the collection layer, but I don’t think it should be implemented there. It should run only when the collection is loaded, which means it belongs in the database layer (i.e., in this class, just as we do with Mongo and JsonFile).

pyproject.toml
guyna25: Changed config to have optional moto with S3 install

src/parlant/adapters/db/s3.py
guyna25: I've adjusted it to be like Mongo and Json. 
I'm unsure if this is the most efficient S3-wise as it offers projection to speed quries, but that will require more config which might be less user friendly. I believe if users use it open source and find it inefficient we will hear about it :)

PR #668, opened 2025-12-10
Perf/optimize snowflake db

Taooo-habitus:
Init collections and drop per query table checks

### DIFF
--- a/examples/healthcare.py
+++ b/examples/healthcare.py
@@ -1,83 +1,8 @@
 # healthcare.py
 
-import asyncio
-from contextlib import AsyncExitStack
 import parlant.sdk as p
-from parlant.adapters.db.snowflake_db import SnowflakeDocumentDatabase
-from parlant.core.emission.event_publisher import EventPublisherFactory
-
-
-# Manage Snowflake-backed stores across the example run
-EXIT_STACK = AsyncExitStack()
-
-
-async def _make_session_store(container: p.Container) -> p.SessionStore:
-    database = await EXIT_STACK.enter_async_context(
-        SnowflakeDocumentDatabase(
-            logger=container[p.Logger],
-            table_prefix="PARLANT_SESSIONS_",
-        )
-    )
-    store = p.SessionDocumentStore(database=database, allow_migration=True)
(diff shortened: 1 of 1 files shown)


src/parlant/adapters/db/snowflake_db.py
```diff
@@ -220,6 +223,93 @@ async def _get_or_create_collection(
 
         return cast(SnowflakeDocumentCollection[TDocument], self._collections[name])
 
+    async def _initialize_collection(
```
mc-dorzo: Do we need both the _get_or_create_collection and the _initialize_collection?

src/parlant/adapters/db/snowflake_db.py
```diff
@@ -220,6 +223,93 @@ async def _get_or_create_collection(
...
+        if name in self._table_ready:
+            return
+
+        lock = self._table_locks.setdefault(name, asyncio.Lock())
+        async with lock:
+            if name in self._table_ready:
+                return
+
+            create_stmt = f"""
+                CREATE TABLE IF NOT EXISTS {collection._table} (
```
mc-dorzo, line 238: This assumes a sessions-like schema. However, for other collections, only id, version, and creation_utc are common. The remaining fields vary by collection, so the table schema must be dynamic.

mc-dorzo: I suggest migrating _get_or_create_collection, _initialize_collection, and _ensure_table together, unless there's a reason to maintain them separated.

Taooo-habitus: You're right! I'll just keep id/version/creation_utc, and everyting else in `data`:

```sql
            INSERT INTO {self._table}
            (ID, VERSION, CREATION_UTC, DATA)
            SELECT
                V.ID,
                V.VERSION,
                V.CREATION_UTC,
                PARSE_JSON(V.DATA_RAW)
            FROM VALUES (
                %(id)s,
                %(version)s,
                %(creation_utc)s,
                %(data)s
            ) AS V(ID, VERSION, CREATION_UTC, DATA_RAW)
```

PR #669, opened 2025-12-11
Fix list of datetime, supporting timedelta

mc-dorzo:


src/parlant/core/engines/alpha/tool_calling/single_tool_batch.py
```diff
@@ -645,13 +645,13 @@ def _add_tool_definitions_section(
         reference_tools: Sequence[tuple[ToolId, Tool]],
     ) -> tuple[str, dict[str, Any]]:
         def _format_type(descriptor_type: str) -> str:
```
mc-dorzo, line 647: move to tool_calling/common.py

PR #680, opened 2025-12-20
Fix the issue where health checks fail due to premature execution when the server is not ready.

itning:
image

If health checks fail due to premature execution.


kichanyurd:
@itning Thanks for pointing this out and for your changeset.

Starting the task later is a problem, as it only introduces a different bug.

Instead, I tried setting the timeout to a higher value. This was just committed here: https://github.com/emcie-co/parlant/commit/d208e78be1d77b8bed62313067ba68251d135c07

Can you please update, try again, and LMK if this fixed your issue?

itning:
Hi @kichanyurd Thank you for your reply. I found that the reason for encountering "Health check failed." was that I had enabled the HTTP proxy. When I turned off the proxy, it worked normally.

BTW: When I enable the proxy, if I immediately trigger an access to this address, the address will return a 502 status code, causing "Health check failed."

PR #685, opened 2025-12-21
Observational_actionable_for_piston

HadarYosef1:
- Fix observational and actionable guideline tests which are not well defined or need to be changed after observational update
- Small fix in backtrack check 
- Fix of bug with tool service registry


MCBarKar:
I noticed something unrelated to this PR while reviewing the matching piston PR - In the output format for actionable guidelines, we do:

`OUTPUT FORMAT
-----------------
- Specify the applicability of each guideline by filling in the details in the following list as instructed:
```json
{{
    {result_structure_text}
}}
```
""",
            props={
                "result_structure_text": self._format_of_guideline_check_json_description(
                    guideline_representations=guideline_representations
                ),
                "guidelines_len": len(self._guidelines),
            },
        )

`


But result_structure_text is already a valid json, so we get two curly brackets instead of one. It's working as is, but we should remove the outer {} from there, both here and in piston. @HadarYosef1 could you please remove this on both PRs?

PR #686, opened 2025-12-22
Improve Python 3.14+ compatible in `plugins` module

KennyDizi:
Problem: In Python `3.14+`, the `Optional[str]` (which is equivalent to `str | None`) wasn't being detected adequately as a union type. The code was checking `generic_type` is `None` to detect union syntax, but in Python 3.14, `types.UnionType` has a `__name__ attribute`, causing the detection to fail.

Fix applied:
1. Added `UnionType` to the imports: `from types import TracebackType, UnionType`
2. Updated the union type detection (line 189) to check for `UnionType` explicitly: `elif get_origin(parameter_type) is UnionType` or `generic_type` is `None`:
3. Fixed the type comparisons `from t is None to t is type(None)` for correctness

### DIFF
--- a/src/parlant/core/services/tools/plugins.py
+++ b/src/parlant/core/services/tools/plugins.py
@@ -324,9 +324,22 @@ def _tool_decorator_impl(
         for param in parameters[1:]:
             param_info = _resolve_param_info(param)
 
-            if issubclass(param_info.resolved_type, enum.Enum):
-                assert all(type(e.value) is str for e in param_info.resolved_type), (
-                    f"{param.name}: {param_info.resolved_type.__name__}: Enum values must be strings"
+            resolved_type = param_info.resolved_type
+            enum_type_to_check: type[enum.Enum] | None = None
+
+            if inspect.isclass(resolved_type) and issubclass(resolved_type, enum.Enum):
+                enum_type_to_check = resolved_type
+            else:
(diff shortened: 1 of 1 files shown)


KennyDizi:
Tested with Python v3.14.2 runtime.

KennyDizi:
@kichanyurd Would you please take a look at this tiny but crucial PR? It makes `Parlant` be able to run with Python v3.14.x

kichanyurd:
@KennyDizi Great catch, thank you!

Can you please run and share the results of the following command, with your fix in place?

`pytest tests/core/stable/services/tools/test_plugin_client.py`

KennyDizi:
Screenshot 2025-12-22 at 16 01 55
Hi @kichanyurd, I've updated the code and run the test again; above is the test result.

kichanyurd:
@KennyDizi great, thank you.

@mc-dorzo can you please rebase locally in a separate branch on top of current develop, and run our tool engine tests?

If everything passes on top of the recent changes we've made there, let's go ahead and merge this PR.

KennyDizi:
thanks @kichanyurd 
Looking forward to it @mc-dorzo

mc-dorzo:
Hey @KennyDizi! Merged the PR — everything looks great.
Thanks for the contribution. Hope to see you here more often… giving me more work 😄

PR #703, opened 2025-12-30
Fix-disambiguation-healthcare

HadarYosef1:


MCBarKar:
Good to merge as far as I'm concerned @mc-dorzo

PR #710, opened 2026-01-05
Remove redundant GLM NLP adapter

majiayu000:
## Summary
- Remove the GLM NLP adapter (`glm_service.py`) as it is superfluous
- The Zhipu adapter already provides the same functionality using the official `zhipuai` SDK
- Remove corresponding `NLPServices.glm()` method from SDK

Closes #697

## Test Plan
- [x] Existing Zhipu adapter tests pass
- [x] Lint checks pass (mypy, ruff)
- [x] No breaking changes - users should migrate to `NLPServices.zhipu()`

PR #719, opened 2026-01-22
fix: Improve LiteLLM service and add embeddings support

avonian:
## Summary

Fixes several issues with the LiteLLM service and adds configurable embeddings support.

**Bug fixes:**
- Use async `acompletion()` instead of blocking `completion()` to avoid blocking the event loop
- Pass `api_key` only if `LITELLM_PROVIDER_API_KEY` is set; otherwise let LiteLLM auto-detect provider-specific keys (`OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, etc.)
- Implement `do_generate()` directly instead of wrapping `_do_generate()` (fixes abstract method error)
- Add `trust_remote_code=True` to HuggingFace model/tokenizer loading to fix JinaAI embeddings with newer transformers versions

**New feature:**
- Add `LiteLLMEmbedder` class that uses `litellm.aembedding()` for provider-agnostic embeddings
- New env var `LITELLM_EMBEDDING_MODEL_NAME` - when set, uses LiteLLM for embeddings; otherwise falls back to local JinaAI
- This allows users to use cloud embedding APIs (OpenAI, Cohere, etc.) without the heavy torch/transformers dependencies

**Note:** This PR supersedes #517 and includes its functionality. The `LITELLM_PROVIDER_BASE_URL` env var (already in the codebase) provides the same self-hosted LLM support that #517 was adding via `LITELLM_PROVIDER_URL`. This PR also addresses the feedback on #517 requesting separate LLM and embedding model configuration.

## Test plan

- [x] Added unit tests for LiteLLMEmbedder initialization and API calls
- [x] Added tests for service embedder selection based on env var

### DIFF
--- a/tests/adapters/nlp/test_litellm_service.py
+++ b/tests/adapters/nlp/test_litellm_service.py
@@ -14,19 +14,19 @@
 
 import asyncio
 import os
-from unittest.mock import AsyncMock, patch, Mock
+from unittest.mock import patch, Mock
 
-import pytest
 from lagom import Container
 
 from parlant.adapters.nlp.litellm_service import (
     LiteLLMEmbedder,
     LiteLLMService,
 )
-from parlant.adapters.nlp.hugging_face import JinaAIEmbedder
 from parlant.core.loggers import Logger
 from parlant.core.meter import Meter
 from parlant.core.tracer import Tracer
+
+import pytest
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


iwr-redmond:
Could tests be accomplished without requiring `trust_remote_code=True`?

avonian:
> Could tests be accomplished without requiring `trust_remote_code=True`?

@iwr-redmond it's needed because Parlant uses JinaAI as the local fallback embedding model, and without `trust_remote_code=True` the Jina model loads as a standard BertModel instead of JinaBertModel (Jina's custom BERT variant).

The custom implementation lives in the model repo and HuggingFace won't execute it without the flag; and because the architectures differ and the checkpoint weights don't map correctly it produces garbage embeddings that break semantic search.

Tests would still technically pass without `trust_remote_code=True`, because they mock the embedder and never actually load the model, but in real use the embeddings would be effectively useless.

I get the security concern of course.

Is it a must to use JinaAI? Because if not, we could replace it with a different sentence-transformers model that works properly (e.g., all-MiniLM-L6-v2, all-mpnet-base-v2, BAAI/bge-small-en-v1.5, BAAI/bge-base-en-v1.5)

src/parlant/adapters/nlp/litellm_service.py
```diff
@@ -125,11 +117,15 @@ async def _do_generate(
             k: v for k, v in hints.items() if k in self.supported_litellm_params
         }
 
+        # Only pass api_key if explicitly set; otherwise let LiteLLM auto-detect
+        # provider-specific keys (OPENAI_API_KEY, ANTHROPIC_API_KEY, etc.)
+        api_key = os.environ.get("LITELLM_PROVIDER_API_KEY") or None
```
mc-dorzo: Why not just 
`api_key = os.environ.get("LITELLM_PROVIDER_API_KEY")`

src/parlant/adapters/nlp/litellm_service.py
```diff
@@ -216,6 +212,64 @@ def max_tokens(self) -> int:
...
+
+    @property
+    @override
+    def tokenizer(self) -> LiteLLMEstimatingTokenizer:
+        return self._tokenizer
+
+    @property
+    @override
+    def max_tokens(self) -> int:
+        return 8192
```
mc-dorzo: Can it be dynamically fetched from an env variable if it exists, with this default value?

src/parlant/adapters/nlp/litellm_service.py
```diff
@@ -216,6 +212,64 @@ def max_tokens(self) -> int:
...
+
+    @property
+    @override
+    def dimensions(self) -> int:
+        # Common dimensions for popular models; may need adjustment per model
+        return 1536
```
mc-dorzo: Also here

src/parlant/adapters/nlp/litellm_service.py
```diff
@@ -216,6 +212,64 @@ def max_tokens(self) -> int:
...
+    def __init__(
+        self,
+        model_name: str,
+        logger: Logger,
+        tracer: Tracer,
+        meter: Meter,
+        base_url: str | None = None,
+    ) -> None:
+        super().__init__(logger, tracer, meter, model_name)
+        self._model_name = model_name
```
mc-dorzo: BaseEmbedder already configures `self.model_name`, so just use that.

src/parlant/adapters/nlp/litellm_service.py
```diff
@@ -216,6 +212,64 @@ def max_tokens(self) -> int:
...
+        return 1536
+
+    @override
+    async def do_embed(
+        self,
+        texts: list[str],
+        hints: Mapping[str, Any] = {},
+    ) -> EmbeddingResult:
+        # Only pass api_key if explicitly set
+        api_key = os.environ.get("LITELLM_PROVIDER_API_KEY") or None
```
mc-dorzo: `api_key = os.environ.get("LITELLM_PROVIDER_API_KEY")`

mc-dorzo:
Overall, looking good mate! Just added some tiny comments.

iwr-redmond:
Off-topic @mc-dorzo but important: you accidentally closed #710 instead of merging it :wink:

avonian:
@mc-dorzo thank you for the feedback, I think everything's been addresses take a look at [4f6caa0](https://github.com/emcie-co/parlant/pull/719/commits/4f6caa017cf794bc6bfe57ef8eddc0be2e3b889f)

standing by

tests/adapters/nlp/test_litellm_service.py
```diff
@@ -0,0 +1,179 @@
+# Copyright 2026 Emcie Co Ltd.
```
kichanyurd, line 1: I'm not sure we need this (I presume generated?) test file.  The tests don't seem particularly meaningful, and ultimately, the maintenance and testing of specific provider adapters is IMO better done manually.

iwr-redmond: Perhaps the LiteLLM tests would be more useful if they tested the environment variables like the [Qwen tests](https://github.com/emcie-co/parlant/blob/faeea1b815ef0942f1c3edf443c52a5a7bd71e21/tests/adapters/nlp/test_qwen_service.py)? LiteLLM is a very flexible service, which means reading the vars is critical for the functionality to work as expected.

avonian: noted, let me see what i can do, will report back

avonian: hey guys, just pushed updated tests, tested quite a bit manually too (setting different env params to change models and embedding models) seems to be working well and jinaAI as fallback embedding model as well

PR #722, opened 2026-01-26
fix: Add CORS headers to exception handler responses

avonian:
## Summary

- Exception handler responses weren't getting CORS headers from FastAPI's CORSMiddleware, causing browsers to block error responses entirely due to missing `Access-Control-Allow-Origin` headers
- Changed exception handlers to return `JSONResponse` instead of raising `HTTPException`
- Added explicit CORS headers via helper function to all error responses (404, 403, 429, 500)
- Fixed incorrect type annotation on `server_error_handler` (was `ItemNotFoundError`, should be `Exception`)

## Test plan

- [x] Added test to verify CORS headers are present on 404 error responses

### DIFF
--- a/src/parlant/api/app.py
+++ b/src/parlant/api/app.py
@@ -20,8 +20,8 @@ from typing import Any, Awaitable, Callable, Mapping, TypeAlias
 
 import mimetypes
 
-from fastapi import APIRouter, FastAPI, HTTPException, Request, Response, status
-from fastapi.responses import RedirectResponse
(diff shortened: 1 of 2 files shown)


kichanyurd:
@avonian Good catch. Can you please rebase on `develop` and remove `uv.lock` from being included in this PR?

avonian:
@kichanyurd all set!

avonian:
just curious if this will get merged, would be nice just so there's one less thing for me to track :grin:

PR #730, opened 2026-02-09
Fix canned response field extraction to handle falsy values

tabajara98:
## Summary

Fixes a bug in `ToolBasedFieldExtraction.extract()` where falsy but valid values (0, False, "", [], {}) set in tool call results were incorrectly skipped due to truthiness evaluation.

## Changes

- Changed the field extraction logic to explicitly check for `None` instead of using truthiness
- Added comprehensive unit tests covering all falsy value types

## Impact

This bug prevented canned response Jinja templates from using conditionals like `{% if result_count == 0 %}` when a tool returned 0 as a valid value. For example, a database query returning 0 results couldn't set `result_count = 0` for template rendering.

## Test Coverage

Added 6 new unit tests in `tests/core/stable/engines/alpha/test_canned_response_field_extraction.py`:
- Integer zero (0)
- Boolean False
- Empty string ("")
- Empty list ([])
- Empty dict ({})
- Verification that None is still returned for missing fields

All tests pass with the fix applied.


kichanyurd:
Hey mate, great catch!

Can you please remove the test file from this PR? I'd like to kind of rethink how we approach the tests here at a later point. For now, it's sufficient that you have indeed run these tests and know that it works.

tabajara98:
Yup, removed it. I just left the fix itself now.

kichanyurd:
Done! Thank you

PR #743, opened 2026-03-07
fix: add ValueError to OpenRouterEmbedder retry policy

### LINKED ISSUE #739: [Bug] Embedding retry policy doesn't cover ValueError: No embedding data received
Body:

  Description

  The @policy retry decorator on OpenRouterEmbedder.do_embed does not retry when the OpenAI
  SDK raises ValueError("No embedding data received"). This causes the entire engine
  preparation iteration to fail on a transient upstream issue.

  Error

  ValueError: No embedding data received

  Raised by the OpenAI SDK's post-parser in openai/resources/embeddings.py when the API
  returns HTTP 200 but with an empty data array.

  Root Cause

  The retry policy in parlant/adapters/nlp/openrouter_service.py (lines ~418-431) only covers
  API transport errors:

  @policy(
      [
          retry(
              exceptions=(
                  APIConnectionError,
                  APITimeoutError,
                  ConflictError,
                  RateLimitError,
                  APIResponseValidationError,
              ),
          ),
          retry(InternalServerError, max_exceptions=2, wait_times=(1.0, 5.0)),
      ]
  )
  async def do_embed(self, texts, hints):

  ValueError is a plain Python exception raised after the HTTP call succeeds, so it bypasses
  the retry policy entirely.

  Impact

  A single transient empty embedding response from the upstream provider kills the entire
  _run_preparation_iteration — glossary terms fail to load and the agent cannot process the
  message.

  ---

kvr06-ai:
## Summary

Fixes #739

The `@policy` retry decorator on `OpenRouterEmbedder.do_embed` only covered API transport errors (`APIConnectionError`, `APITimeoutError`, etc.) and `InternalServerError`. When the upstream provider returns HTTP 200 with an empty data array, the OpenAI SDK raises `ValueError("No embedding data received")` in its post-parser. Because `ValueError` is a plain Python exception, it bypassed the retry policy entirely, causing the full `_run_preparation_iteration` to fail on a transient issue.

This PR adds `ValueError` to the embedder's retry policy with conservative limits (max 2 retries, 1s/5s backoff), matching the existing `InternalServerError` retry entry.

## Changes

- `src/parlant/adapters/nlp/openrouter_service.py`: Added `retry(ValueError, max_exceptions=2, wait_times=(1.0, 5.0))` to the `do_embed` policy decorator
- `tests/adapters/nlp/test_openrouter_service.py`: Added regression test that verifies `do_embed` retries on `ValueError` and succeeds on a subsequent call

## Testing

- New test `test_that_openrouter_embedder_retries_on_valueerror` mocks a `ValueError` on the first call followed by a successful response, then asserts the embedder retries and returns correct results
- Full test suite passes (26 passed, 1 skipped)

### DIFF
--- a/src/parlant/adapters/nlp/openrouter_service.py
+++ b/src/parlant/adapters/nlp/openrouter_service.py
@@ -432,6 +432,7 @@ class OpenRouterEmbedder(BaseEmbedder):
                 ),
             ),
             retry(InternalServerError, max_exceptions=2, wait_times=(1.0, 5.0)),
+            retry(ValueError, max_exceptions=2, wait_times=(1.0, 5.0)),
         ]
     )
     @override
--- a/tests/adapters/nlp/test_openrouter_service.py
+++ b/tests/adapters/nlp/test_openrouter_service.py
@@ -25,6 +25,7 @@ from openai.types.completion_usage import CompletionUsage
 from parlant.adapters.nlp.openrouter_service import (  # type: ignore[reportMissingImports]
     OpenRouterService,
     OpenRouterSchematicGenerator,
+    OpenRouterEmbedder,
     OpenRouterGPT4O,
     OpenRouterGPT4OMini,
     OpenRouterClaude35Sonnet,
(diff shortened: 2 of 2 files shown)


kvr06-ai:
@mc-dorzo Would you mind taking a look when you get a chance? Small one-liner fix with a regression test.

santangelx:
I believe this was fixed [here](https://github.com/emcie-co/parlant/pull/746)

kvr06-ai:
Closing in favor of #746, which already addressed this.

PR #744, opened 2026-03-07
test: add regression test for embedding cache eviction (#731)

### LINKED ISSUE #731: [Bug] KeyError in LRU embedding cache eviction when entries share the same text length
# Description
BaseEmbedder._cache_put can raise a KeyError during cache eviction when two or more cached texts share the same character length. The eviction logic unconditionally removes the text_length key from _cache_length_index, even when other entries with that same length are still present in the cache.

# Location
src/parlant/core/nlp/embedding.py, lines 162-166:

```python
if len(self._cache) >= _EMBEDDING_CACHE_MAX_SIZE:
    oldest_checksum, oldest_entry = self._cache.popitem(last=False)
    self._cache_length_index[oldest_entry.text_length].discard(oldest_checksum)
    self._cache_length_index.pop(oldest_entry.text_length, None)
```

# How to Reproduce
1. The cache reaches capacity (_EMBEDDING_CACHE_MAX_SIZE = 1000)
2. Two entries share the same text_length, e.g. 592:
    * Entry A (checksum=100, oldest) and Entry B (checksum=200)
    * _cache_length_index[592] = {100, 200}
3. A new text triggers eviction, and Entry A is popped:
    * .discard(100) updates the set to {200}
    * .pop(592, None) removes the key 592 entirely, orphaning Entry B
4. Eventually Entry B becomes the oldest and gets evicted:
    * self._cache_length_index[592] raises KeyError: 592

# Environment
 - Docker linux
 - Python 3.12
 - parlant[mongo]~=3.1.2
 - parlant-client~=3.1.0

kvr06-ai:
## Summary

Adds a regression test for the `KeyError` in `BaseEmbedder._cache_put` when two cached texts share the same character length (reported in #731). The code fix is already on `develop`; this PR adds the missing test coverage the maintainer requested.

Fixes #731

## Changes

- New file `tests/core/test_embedding_cache.py` with two test cases:
  - `test_that_cache_eviction_does_not_raise_when_entries_share_text_length`: reproduces the exact scenario from #731 (two entries share `text_length`, evicting the older one must not orphan the newer one's length-index entry)
  - `test_that_cache_eviction_cleans_up_length_index_for_unique_lengths`: verifies the length-index key is properly removed when the last entry with that length is evicted

## Testing

- Both tests pass on `develop` (with the fix in place)
- Verified the primary test fails with the original buggy code (`KeyError: 3` at `_cache_put` line 165), confirming it catches the regression
- Uses `unittest.mock.patch` to set `_EMBEDDING_CACHE_MAX_SIZE = 3` so tests run in milliseconds without needing 1000 cache entries

### DIFF
--- a/tests/core/test_embedding_cache.py
+++ b/tests/core/test_embedding_cache.py
@@ -1,0 +1,119 @@
+# Copyright 2026 Emcie Co Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from collections.abc import Mapping
+from typing import Any
+from unittest.mock import MagicMock, patch
+
+from typing_extensions import override
+
+from parlant.core.nlp.embedding import BaseEmbedder, EmbeddingResult
+from parlant.core.nlp.tokenization import ZeroEstimatingTokenizer, EstimatingTokenizer
+
+
+class _StubEmbedder(BaseEmbedder):
+    """Minimal embedder for testing cache behavior."""
+
+    @override
+    async def do_embed(
+        self,
+        texts: list[str],
+        hints: Mapping[str, Any] = {},
+    ) -> EmbeddingResult:
+        return EmbeddingResult(vectors=[[0.0] for _ in texts])
+
+    @property
+    @override
+    def id(self) -> str:
+        return "stub"
+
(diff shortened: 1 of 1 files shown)


kvr06-ai:
@kichanyurd This adds the regression test for the cache eviction bug you flagged. The test reproduces the exact scenario from the issue (two entries sharing `text_length`, sequential eviction) and confirms it raises `KeyError: 3` without the fix. Happy to adjust if you'd prefer a different test structure or location.

kichanyurd:
@kvr06-ai Hey there, we've actually added the fix already - it's on the `develop` branch. But it would be great if you could verify it works on your app. Could you please LMK?

kvr06-ai:
Yes, verified on Python 3.12. I reproduced the exact eviction sequence from the issue in a unit test: with the original code it raises `KeyError: 3`, with the fix on `develop` it passes cleanly. The bug is pure in-memory Python logic, so the environment (Docker, mongo adapter) doesn't affect it.

kichanyurd:
@kvr06-ai Excellent! Thank you for finding and reporting this important issue - and glad to hear it now works as expected.

PR #745, opened 2026-03-07
Improve session persistence hot paths

santangelx:
## Summary

This PR improves Parlant's session persistence hot paths in two ways:

1. `SessionDocumentStore.create_event()` no longer lists every visible event in a session just to compute the next offset.
2. Mongo-backed session and customer stores now register indexes for the query patterns they already rely on.

The event-offset change also fixes a correctness bug: deleted events were excluded from the offset calculation, so creating a new event after deleting the latest one could reuse an existing offset.

## Changes

- add a sorted `find_one()` capability to document collections so stores can fetch the latest matching document efficiently
- add collection-level `ensure_indexes()` support, with MongoDB creating the requested indexes and other document backends treating it as a no-op
- allocate new event offsets from the highest existing session offset, including deleted events
- register Mongo indexes for:
  - session lookup by `id`
  - session listings by `agent_id` / `customer_id` with `creation_utc` pagination
  - event lookup by `id`
  - per-session event scans by `session_id` and `offset`
  - customer lookup by `id`
  - customer tag association lookups by `customer_id` / `tag_id`
- add a regression test that verifies offsets keep increasing after deleted events
- add Mongo integration tests that assert the session and customer stores provision the expected indexes

## Tests

Ran locally:

- `ruff check src/parlant/core/persistence/document_database.py src/parlant/adapters/db/mongo_db.py src/parlant/adapters/db/transient.py src/parlant/adapters/db/json_file.py src/parlant/adapters/db/snowflake_db.py src/parlant/core/sessions.py src/parlant/core/customers.py tests/api/test_sessions.py tests/adapters/db/test_mongodb.py`
- `python -m py_compile src/parlant/core/persistence/document_database.py src/parlant/adapters/db/mongo_db.py src/parlant/adapters/db/transient.py src/parlant/adapters/db/json_file.py src/parlant/adapters/db/snowflake_db.py src/parlant/core/sessions.py src/parlant/core/customers.py tests/api/test_sessions.py tests/adapters/db/test_mongodb.py`
- `pytest tests/api/test_sessions.py -q -k "deleted_events_no_longer_show_up_in_the_listing or keep_increasing_offsets_after_deleted_events"`
- `pytest tests/adapters/db/test_json_file.py -q -k "dummy_documents_can_be_created or dummy_documents_can_be_read_by_id"`
- `pytest tests/adapters/db/test_mongodb.py -q -k "session_store_creates_indexes or customer_store_creates_indexes"`

### DIFF
--- a/src/parlant/core/sessions.py
+++ b/src/parlant/core/sessions.py
@@ -918,6 +918,12 @@ class SessionDocumentStore(SessionStore):
                     CollectionIndex(fields=(("id", SortDirection.ASC),)),
                     CollectionIndex(
                         fields=(
+                            ("creation_utc", SortDirection.ASC),
+                            ("id", SortDirection.ASC),
+                        )
+                    ),
+                    CollectionIndex(
+                        fields=(
                             ("agent_id", SortDirection.ASC),
                             ("creation_utc", SortDirection.ASC),
                             ("id", SortDirection.ASC),
--- a/src/parlant/core/sessions.py
+++ b/src/parlant/core/sessions.py
@@ -920,14 +920,14 @@ class SessionDocumentStore(SessionStore):
                         fields=(
                             ("agent_id", SortDirection.ASC),
                             ("creation_utc", SortDirection.ASC),
-                            ("_id", SortDirection.ASC),
+                            ("id", SortDirection.ASC),
                         )
                     ),
                     CollectionIndex(
                         fields=(
                             ("customer_id", SortDirection.ASC),
                             ("creation_utc", SortDirection.ASC),
-                            ("_id", SortDirection.ASC),
+                            ("id", SortDirection.ASC),
                         )
                     ),
                 ]
@@ -1316,10 +1316,7 @@ class SessionDocumentStore(SessionStore):
             creation_utc = creation_utc or datetime.now(timezone.utc)
             latest_event = await self._event_collection.find_one(
                 filters={"session_id": {"$eq": session_id}},
-                sort=(
-                    ("offset", SortDirection.DESC),
-                    ("id", SortDirection.DESC),
-                ),
+                sort=(("offset", SortDirection.DESC),),
             )
             offset = latest_event["offset"] + 1 if latest_event else 0
 
--- a/src/parlant/core/sessions.py
+++ b/src/parlant/core/sessions.py
@@ -56,6 +56,7 @@ from parlant.core.persistence.common import (
 )
 from parlant.core.persistence.document_database import (
     BaseDocument,
+    CollectionIndex,
     DocumentDatabase,
     DocumentCollection,
 )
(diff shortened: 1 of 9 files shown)


kichanyurd:
@mc-dorzo Could you please review this one? Looks highly relevant

src/parlant/core/sessions.py
```diff
@@ -912,6 +913,43 @@ async def __aenter__(self) -> Self:
...
                 document_loader=self._event_document_loader,
             )
+            await self._session_collection.ensure_indexes(
+                [
+                    CollectionIndex(fields=(("id", SortDirection.ASC),)),
+                    CollectionIndex(
+                        fields=(
+                            ("agent_id", SortDirection.ASC),
+                            ("creation_utc", SortDirection.ASC),
+                            ("_id", SortDirection.ASC),
```
mc-dorzo: "_id" is implementation detail for Mongo, rename it to "id"

src/parlant/core/sessions.py
```diff
@@ -1275,11 +1313,15 @@ async def create_event(
...
-            session_events = await self.list_events(
-                session_id
-            )  # FIXME: we need a more efficient way to do this
             creation_utc = creation_utc or datetime.now(timezone.utc)
-            offset = len(list(session_events))
+            latest_event = await self._event_collection.find_one(
+                filters={"session_id": {"$eq": session_id}},
+                sort=(
+                    ("offset", SortDirection.DESC),
+                    ("id", SortDirection.DESC),
```
mc-dorzo: The `id` is uuid4 / nanoid, not an ascending ID. Therefore sorting by it is not correct. We also assume events have different ascending offsets within a session, so that should be sufficient.

src/parlant/core/sessions.py
mc-dorzo: creation_utc need to be first before agent_id

mc-dorzo:
Very nice PR, mate!
I left a few comments. Please also run uv run mypy src/ and uv run mypy tests/ and fix any errors if you find them.

src/parlant/core/sessions.py
santangelx: About the index order : since our query lists sessions by getting an agent/customer and then orders by time, we should keep the current order.

src/parlant/core/sessions.py
santangelx: Good catch!

src/parlant/core/sessions.py
mc-dorzo: Good point about the query pattern! However, currently agent/customer are optional, and we also need to support platform-level queries that list all sessions, regardless of agent/customer (both fields are optional). In those cases, the query filters only by creation_utc range, and having it as the first index field allows MongoDB to use the index for these broader queries efficiently.

With creation_utc first, the index serves both use cases:
Platform queries: Filter by time range across all sessions → uses index prefix
Agent/customer queries: Filter by agent/customer + time → still works, just scans the time-sorted entries

If agent_id is first, platform-wide time-range queries can't leverage the index efficiently since there's no leading equality match.

What do you think?

santangelx: You’re right that unfiltered session listing exists, but that’s a different query pattern. Reordering the per-agent/per-customer indexes would hurt the hot filtered path and still wouldn’t be the best global index. If we want to optimize global listing too, the correct additive index is (creation_utc, id), not creation_utc-first agent/customer compounds. We could add that on another PR

mc-dorzo: That's a great idea. I would love to add this to this PR, though, since I see it as critical.

mc-dorzo: That's a great idea. I would love to add this to this PR, though, since I see it as critical.

santangelx: Done!

mc-dorzo:
Merged! Good job mate. see you on the next PR ;)

PR #747, opened 2026-03-08
Fix canned response vector document IDs

santangelx:
## What changed
- give each canned response vector document its own ID
- keep the existing multi-vector behavior for `value` plus `signals`
- add a regression test that covers repeated signal text and value overlap

## Why
Persistent vector backends key documents by ID. Reusing the canned response ID for every vector means later inserts can stomp earlier ones, which breaks the whole multi-vector setup.

## Tests
- `PYTHONPATH=src python -m pytest tests/core/stable/test_canned_response_vector_store.py -q`
- tried `PYTHONPATH=src python -m pytest tests/api/test_canned_responses.py -x -vv`
- tried `PYTHONPATH=src python -m pytest tests/sdk/test_canned_responses.py -x -vv`

The API and SDK canned-response suites are blocked locally by the default test embedder setup requiring `EMCIE_API_KEY`.

### DIFF
--- a/tests/core/stable/test_canned_response_vector_store.py
+++ b/tests/core/stable/test_canned_response_vector_store.py
@@ -1,14 +1,42 @@
-from typing import cast
+# Copyright 2026 Emcie Co Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
-import pytest
+from collections.abc import Mapping
+from typing import Any, cast
+
 from lagom import Container
+import pytest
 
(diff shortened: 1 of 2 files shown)


tests/core/stable/test_canned_response_vector_store.py
```diff
@@ -0,0 +1,33 @@
...
+    container: Container,
+) -> None:
+    store = cast(CannedResponseVectorStore, container[CannedResponseStore])
+    inserted_documents: list[CannedResponseVectorDocument] = []
+
+    async def capture_insert(document: CannedResponseVectorDocument) -> InsertResult:
+        inserted_documents.append(document)
+        return InsertResult(acknowledged=True)
+
+    store._canreps_vector_collection.insert_one = capture_insert  # type: ignore[method-assign]
```
mc-dorzo: Do not use private variables. Insert multiple canreps with the same content and retrieve back and see the result as expected.
Its also won't pass mypy

santangelx: Sorry about that, redid it properly.

santangelx:
Rebased onto latest develop (clean, no conflicts).

Testing done:
- `uv run pytest tests/core/stable/test_canned_response_vector_store.py -v` — 2 tests pass
- `uv run pytest tests/api/test_canned_responses.py -v` — 21 tests pass
- `uv run pytest tests/sdk/test_canned_responses.py -v` — 2 tests pass

API and SDK suites now run fine with the EMCIE key.

PR #749, opened 2026-03-08
Keep capability vectors in sync

santangelx:
## What changed
- remove old vector docs before re-inserting capability vectors on update
- remove capability vector docs on delete too
- add regression coverage around update and delete behavior

## Why
Without the cleanup, capability vectors can stick around after updates or deletes and keep showing up in retrieval.

## Tests
- `PYTHONPATH=src python -m pytest tests/core/stable/test_capability_vector_store.py -q`

I kept the tests self-contained with a tiny local embedder stub so they don't depend on external credentials.

### DIFF
--- a/src/parlant/core/capabilities.py
+++ b/src/parlant/core/capabilities.py
@@ -308,6 +308,18 @@ class CapabilityVectorStore(CapabilityStore):
 
         return doc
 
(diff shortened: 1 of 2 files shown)


mc-dorzo:
Merged! great catch mate

PR #750, opened 2026-03-08
Refresh journey vectors after graph edits

santangelx:
## What changed
- refresh the stored journey vector when nodes or edges are created, updated, or deleted
- add regression tests that check the vector content stays in sync with graph edits

## Why
Journey embeddings are built from the journey title/description plus node actions and edge conditions. If those graph bits change and the vector doesn't, retrieval gets stale fast.

## Notes
I left journey conditions out on purpose here, since they are not part of the embedded content today.

## Tests
- `PYTHONPATH=src python -m pytest tests/core/stable/test_journey_vector_store.py -q`

These tests also use a local embedder stub so they stay runnable without external API keys.


mc-dorzo:
Merged! Nice one

PR #751, opened 2026-03-09
Fix MCP service teardown when replacing services

santangelx:
A small one, but it fixes a real lifecycle bug in the MCP path.

What changed:
- remove the duplicate `enter_async_context()` call in `ServiceDocumentRegistry.update_tool_service()`
- add a regression test that proves the previous MCP client is actually disconnected after a service update

Why this matters:
The registry was entering remote services twice and only manually exiting them once on update/delete. For MCP clients that means the old connection could stay half-alive instead of being fully torn down, which makes recovery and replacement a lot messier than it should be.

Testing:
- `PYTHONPATH=src /Users/santangelo/projects/parlant/.venv/bin/pytest tests/core/stable/engines/alpha/test_mcp.py -k "updating_an_mcp_service_closes_the_previous_client_connection or reading_an_existing_mcp_service_returns_its_tools_and_can_call_them"`

### DIFF
--- a/src/parlant/core/services/tools/service_registry.py
+++ b/src/parlant/core/services/tools/service_registry.py
@@ -332,12 +332,6 @@ class ServiceDocumentRegistry(ServiceRegistry):
 
             self._running_services[name] = service
 
-        await self._exit_stack.enter_async_context(
-            self._cast_to_specific_tool_service_class(service)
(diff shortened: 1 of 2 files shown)


mc-dorzo:
Good catch mate!

PR #752, opened 2026-03-09
Preserve MCP endpoints across service reloads

santangelx:
This one tightens up MCP service persistence so reloads don't quietly point at the wrong place.

What changed:
- keep the full MCP endpoint when serializing services, instead of dropping the port
- return the full stored endpoint from the API layer
- accept the current service document version when loading persisted service records
- add a roundtrip persistence test for an MCP service on a custom port

Why this matters:
If an MCP service was registered on anything other than the default port, Parlant could persist only the host and then rebuild the client with the wrong endpoint later. On top of that, the registry loader was rejecting its own current service document version on reload. This PR makes the saved endpoint survive the roundtrip cleanly.

Testing:
- `PYTHONPATH=src /Users/santangelo/projects/parlant/.venv/bin/pytest tests/core/stable/engines/alpha/test_mcp.py -k "endpoint_roundtrips_through_persistence or reading_an_existing_mcp_service_returns_its_tools_and_can_call_them"`
- `PYTHONPATH=src /Users/santangelo/projects/parlant/.venv/bin/pytest tests/api/test_services.py -k "mcp_service_is_created or reading_an_existing_mcp_service_returns_its_metadata_and_tools"`

### DIFF
--- a/src/parlant/core/services/tools/service_registry.py
+++ b/src/parlant/core/services/tools/service_registry.py
@@ -259,7 +259,7 @@ class ServiceDocumentRegistry(ServiceRegistry):
             url = service.url
         elif isinstance(service, MCPToolClient):
             kind = "mcp"
-            url = service.url
+            url = service.endpoint_url
         else:
             raise ValueError("Unsupported ToolService class.")
 
--- a/src/parlant/api/services.py
+++ b/src/parlant/api/services.py
@@ -251,7 +251,7 @@ def _get_service_url(service: ToolService) -> str:
     if isinstance(service, PluginClient):
         return service.url
     if isinstance(service, MCPToolClient):
-        return f"{service.url}:{service.port}"
+        return service.endpoint_url
     raise ValueError(f"Unknown service kind: {type(service)}")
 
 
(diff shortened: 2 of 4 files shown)


santangelx:
Rebased onto latest develop and resolved conflicts.

Testing done:
- `uv run pytest tests/core/stable/engines/alpha/test_mcp.py -v` — all 14 tests pass (including new persistence roundtrip test)
- `uv run pytest tests/api/test_services.py -v` — all 12 tests pass

src/parlant/core/services/tools/service_registry.py
```diff
@@ -163,7 +163,7 @@ def _cast_to_specific_tool_service_class(
         return service
 
     async def _document_loader(self, doc: BaseDocument) -> Optional[_ToolServiceDocument]:
-        if doc["version"] == "0.1.0":
+        if doc["version"] in {"0.1.0", self.VERSION.to_string()}:
```
mc-dorzo: Good job you found a bug! I just fixed it in develop. please rebase and drop this change :)

PR #753, opened 2026-03-09
Reconnect MCP clients after stale sessions

santangelx:
This one is about making the MCP wrapper less brittle when the underlying FastMCP session goes stale.

What changed:
- initialize and manage the FastMCP client through a single lock-protected connection path
- recreate the client when the current session is no longer connected
- retry initial MCP connection attempts a few times before giving up
- retry metadata reads once after reconnect when the failure looks like a dropped session
- add tests for stale-session recovery and initial connection retry

Why this matters:
Right now once the wrapped FastMCP client falls out of a connected state, Parlant just keeps talking to the same dead client and MCP calls fail from there. This PR gives the wrapper a way to heal itself instead of staying broken until the whole service object is replaced.

Testing:
- `PYTHONPATH=src /Users/santangelo/projects/parlant/.venv/bin/pytest tests/core/stable/engines/alpha/test_mcp.py`

### DIFF
--- a/src/parlant/core/services/tools/mcp_service.py
+++ b/src/parlant/core/services/tools/mcp_service.py
@@ -20,9 +20,10 @@ import json
 from mailbox import FormatError
 from mcp.types import Tool as McpTool
 from types import TracebackType
-from typing import Any, Sequence, Mapping, Optional, Literal, Callable, cast
+from typing import Any, Sequence, Mapping, Optional, Literal, Callable, Awaitable, cast
 from typing_extensions import override
 import asyncio
+import httpx
 
 from fastmcp import FastMCP
 from fastmcp.tools import Tool as FastMCPTool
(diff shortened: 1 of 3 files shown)


mc-dorzo:
I see many mypy errors from another PR we merged regarding MCP. Please fix them in this PR.
 
 
image

santangelx:
Rebased onto latest develop and resolved conflicts.

Testing done:
- `uv run pytest tests/core/stable/engines/alpha/test_mcp.py tests/core/stable/services/tools/test_mcp_client.py -v` — all 15 tests pass
- Fixed 3 mypy `union-attr` errors on `_client` in test assertions

mc-dorzo:
Can you please rebase develop and resolve conflicts?

santangelx:
@mc-dorzo done :)

mc-dorzo:
Merged! Thank @santangelx !

PR #770, opened 2026-04-06
fix(api): preserve falsy values in session patch

### LINKED ISSUE #768: Session PATCH silently drops falsy values (client=0, title=\"\", empty metadata ops)
## Bug Description

In `src/parlant/api/sessions.py`, the session PATCH endpoint uses truthy checks (`if dto.xxx:`) instead of `if dto.xxx is not None:` for optional fields. This silently drops updates when the new value is falsy — e.g. `client=0`, `title=""`, or `metadata={"set": {}}`.

```python
# Current code — all of these skip falsy values:
if dto.consumption_offsets:        # skips when client=0
if dto.title:                      # skips empty string
if dto.mode:                       # would skip if mode had a falsy variant
if dto.metadata:                   # skips empty metadata ops
if dto.metadata.set:               # skips empty dict
if dto.metadata.unset:             # skips empty list
```

The same pattern also exists in the event metadata update path (around line 2236).

## Steps to Reproduce

```python
# PATCH a session with client offset = 0
response = client.patch(f"/sessions/{sid}", json={"consumption_offsets": {"client": 0}})
# Expected: consumption_offsets.client set to 0
# Actual: update silently skipped, offset unchanged
```

## Expected Behavior

All `is not None` semantics — falsy values like `0`, `""`, `{}`, `[]` should be accepted as valid updates.

## Environment

- Branch: `develop` (commit `bee7790`)

YizukiAme:
Fixes #768

Session PATCH was using truthy checks (`if dto.xxx:`) for optional fields, which silently drops valid falsy values like `client=0`, `title=""`, or empty metadata operations.

Changed all 7 checks to `is not None` so that falsy values are accepted as valid updates. Metadata `set`/`unset` inner guards are kept as truthy checks since empty dict/list operations are already no-ops.

Added tests covering `client=0`, `title=""`, and omitting `consumption_offsets` entirely.


src/parlant/api/sessions.py
```diff
@@ -1608,35 +1608,35 @@ async def update_session(
...
+            if dto.agent_id is not None:
                 params["agent_id"] = dto.agent_id
 
-            if dto.metadata:
+            if dto.metadata is not None:
                 session = await app.sessions.read(session_id)
                 current_metadata = dict(session.metadata)
 
-                if dto.metadata.set:
+                if dto.metadata.set is not None:
```
mc-dorzo: dto.metadata.set is either a dictionary or None. If it’s not None but falsy, it’s simply an empty dictionary, and current_metadata.update(...) will be a no-op. Therefore, this check (and the same for unset) is redundant and can be removed.

src/parlant/api/sessions.py
```diff
@@ -2236,9 +2236,9 @@ async def update_event(
         if params.metadata is not None:
             # Convert API DTO to app_modules model - pass through set/unset operations
             metadata_update: EventMetadataUpdateParamsModel = {}
-            if params.metadata.set:
+            if params.metadata.set is not None:
```
mc-dorzo: Also here

YizukiAme:
Apologies for the mess... accidentally ran a force-push from a shallow clone, which corrupted the commit history and inflated the diff to 766 files. That's what caused the PR to break.
I've already addressed @mc-dorzo's review feedback (reverted the redundant is not None guards on metadata.set/unset back to truthy checks). 
Will open a clean PR with the corrected branch shortly!
Again — Sorry for the noise!😇

PR #771, opened 2026-04-06
fix(sessions): avoid agent state truncation crash on non-agent traces

### LINKED ISSUE #769: delete_events() crashes with StopIteration on non-agent traces
## Bug Description

`SessionModule.delete_events()` in `src/parlant/core/app_modules/sessions.py` crashes with `StopIteration` when the event being deleted belongs to a non-agent trace (e.g. a human message trace).

The crash happens at the `next()` call that looks for a matching `agent_state`:

```python
state_index_offset = next(
    i
    for i, s in enumerate(session.agent_states, start=0)
    if s.trace_id.startswith(event_at_min_offset.trace_id)
)
```

When the deleted event's `trace_id` has no corresponding entry in `session.agent_states` (which is the case for human-originated events), the generator is empty and `next()` raises `StopIteration`. This propagates as an unhandled exception.

Additionally, the function does not sort events by offset before selecting the cutoff event, which could lead to incorrect state truncation if events are returned in non-deterministic order.

## Steps to Reproduce

1. Create a session with at least one human message event
2. Call `delete_events()` targeting that human event's offset
3. `StopIteration` is raised

## Expected Behavior

- Deleting a non-agent trace should succeed and leave `agent_states` unchanged
- Events should be sorted by offset before selecting the cutoff point

## Environment

- Branch: `develop` (commit `bee7790`)

YizukiAme:
Fixes #769

`delete_events()` crashes with `StopIteration` when the cutoff event belongs to a human trace — the `next()` call looking for a matching `agent_state` finds nothing because human traces have no agent state entry.

Fixed by passing a default to `next(..., None)` and returning early when no matching agent state exists. Also added `sorted()` by offset before selecting the cutoff event to avoid issues with non-deterministic event ordering.

Added a human-trace regression test and tightened the existing agent-trace test to select by `trace_id`.

### DIFF
--- a/src/parlant/core/app_modules/sessions.py
+++ b/src/parlant/core/app_modules/sessions.py
@@ -517,10 +517,13 @@ class SessionModule:
     ) -> None:
         session = await self._session_store.read_session(session_id)
 
-        events = await self._session_store.list_events(
-            session_id=session_id,
-            min_offset=0,
-            exclude_deleted=True,
+        events = sorted(
+            await self._session_store.list_events(
+                session_id=session_id,
+                min_offset=0,
+                exclude_deleted=True,
+            ),
+            key=lambda event: event.offset,
         )
 
         events_starting_from_min_offset = [e for e in events if e.offset >= min_offset]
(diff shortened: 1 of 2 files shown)


mc-dorzo:
Nice catch mate. merged!

PR #785, opened 2026-04-24
fix: handle bracketless list arguments from LLM

### LINKED ISSUE #431: [Bug] Deal with list arguments that are provided by LLM without brackets

octo-patch:
Fixes #431

## Problem

LLMs occasionally produce list arguments as comma-separated values **without surrounding brackets** (e.g. `red, green` instead of `['red', 'green']`). This non-deterministic hallucination caused crashes in:

- `split_arg_list()` in `core/tools.py` — raised `ValueError` (invalid list format)
- `split_arg_list()` in `core/services/tools/mcp_service.py` — raised `ValueError`
- `_validate_argument_value()` in both tool-calling batch modules — `ast.literal_eval()` raised `SyntaxError` / `ValueError`

## Solution

- **`split_arg_list()` (str/Enum item types)**: Try `literal_eval()` first; fall back to comma-splitting with strip when it fails.
- **`split_arg_list()` (scalar types like int/float)**: Strip optional `[...]` brackets if present, then split by comma — making brackets optional.
- **`_validate_argument_value()`** (both batch files): Wrap `ast.literal_eval()` in a try/except and fall back to comma-splitting for array enum validation.

## Testing

Added 13 unit tests in `tests/core/stable/test_tool_argument_handling.py` covering:
- Already-a-list passthrough
- Bracketed and bracketless string lists
- Bracketed and bracketless enum lists (with and without spaces)
- Bracketed and bracketless int lists
- `cast_tool_argument` with bracketless `list[int]`, `list[str]`, and `list[Enum]`

All 13 tests pass. Ruff lint and mypy checks pass (pre-existing mypy errors in optional adapter stubs are unrelated to this change).


kichanyurd:
Hi @octo-patch - thanks for the contribution! Can you please resolve the conflicts and push?

PR #794, opened 2026-05-10
Reject local filesystem paths in OpenAPI service source (CWE-22)

sebastionoss:
## Summary

The `PUT /services/{name}` and `PATCH /services/{name}` endpoints (OpenAPI service kind) accept an arbitrary string as `openapi.source`. When the value is not an `http(s)` URL, `ServiceDocumentRegistry._get_openapi_json_from_source` falls through to `aiofiles.open(source)` and reads the path from the local filesystem. A caller able to reach the services API can therefore have the server read any file the process has access to (e.g. `/etc/passwd`, app config, secrets exposed via `/proc/self/environ`). The contents are then handed to `openapi_parser`, where parse errors and the resulting tool metadata can leak file content back via API responses and logs.

- CWE: [CWE-22 — Improper Limitation of a Pathname to a Restricted Directory](https://cwe.mitre.org/data/definitions/22.html) (with CWE-73 elements — externally controlled file path)
- Severity: High in default deployments
- Affected files: `src/parlant/api/services.py`, `src/parlant/core/services/tools/service_registry.py`

### Data flow

1. `services.py::create_or_update_service` accepts the request body and stores `openapi.source` verbatim.
2. `service_registry.py::ServiceDocumentRegistry.update_tool_service` calls `_get_openapi_json_from_source(source)`.
3. That helper checks `source.startswith("http://") or source.startswith("https://")`; otherwise it does `aiofiles.open(source, "r") as file: return await file.read()`.
4. The body is parsed by `openapi_parser`; parse errors propagate through the API response, and successfully parsed tool descriptions become listable via the services endpoints.

### Preconditions

- Network reachability to `parlant-server`.
- `DevelopmentAuthorizationPolicy` active (the default unless `PARLANT_ENV=production`), which permits all operations and configures CORS `allow_origins=['*']` with credentials. A custom policy that grants `Operation.UPDATE_SERVICE` to a low-privileged caller has the same effect.

## Fix

Two layers of defense, kept intentionally small:

1. **API boundary (`api/services.py`)** — validate that `openapi.source` begins with `http://` or `https://` and return HTTP 422 otherwise. This is the primary fix: untrusted input never reaches the filesystem.
2. **Core registry (`core/services/tools/service_registry.py`)** — the file-based branch of `_get_openapi_json_from_source` is now gated behind the `PARLANT_ALLOW_OPENAPI_FILE_SOURCE` env var. This preserves backward compatibility for operators who legitimately persisted file paths from prior installs, but only when they explicitly opt in.

Why both layers? The registry is reachable from the SDK as well as the HTTP API, so a defense-in-depth check there also protects callers that bypass the FastAPI layer (e.g. embedded use of the registry).

## Tests

- Replaced the previous "service can be added via file" e2e test (`tests/e2e/test_client_cli_via_api.py`) with `test_that_an_openapi_service_added_via_file_is_rejected`, which asserts the CLI exits non-zero and the service is not created.
- Updated `tests/api/test_services.py` to assert HTTP 422 on local-path sources and to keep covering the `http(s)` happy path.
- Local run: existing service test suite passes; the new negative test fails on `develop` and passes on this branch.

## Adversarial review

Before submitting, we tried to disprove this. The main thing to rule out was whether the default authorization policy already granted equivalent access — it doesn't: `DevelopmentAuthorizationPolicy` allows API calls but does not give the caller arbitrary file read on the host, and CORS `allow_origins=['*']` actually widens the attack surface (any origin can drive a victim browser at a dev instance). We also checked whether `openapi_parser` would reject non-OpenAPI content early enough to prevent disclosure — it does in many cases, but error messages and the parsed-tool metadata still leak content for files that happen to be valid YAML/JSON, and the read itself is observable via timing/error channels regardless. Finally, we considered whether validating only at the API would be sufficient; we kept the registry-level gate because the registry is also reachable from in-process SDK callers.

## Notes for maintainers

- No public API removed. `openapi.source` semantics are unchanged for `http(s)` URLs.
- Operators who deliberately serve OpenAPI specs from local files can re-enable the old behavior by setting `PARLANT_ALLOW_OPENAPI_FILE_SOURCE=1`. This is documented in `CHANGELOG.md`.
- Happy to split into two commits or adjust the env-var name if you prefer a different convention.

_Submitted by Sebastion — autonomous open-source security research from [Foundation Machines](https://foundationmachines.ai). Free for public repos via the [Sebastion AI GitHub App](https://github.com/marketplace/sebastion-ai)._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to Parlant will be documented here.
 
 ## [Unreleased]
 
+### Security
+
+- Reject local filesystem paths in the OpenAPI tool service `source` field at the API boundary, and require operators to opt in via `PARLANT_ALLOW_OPENAPI_FILE_SOURCE=true` to load OpenAPI specs from disk in the registry. Prevents arbitrary local file reads via `PUT /services/{name}` (CWE-22).
+
 ### Added
 
 - Add `AnyOf(tag)` and `AllOf(tag)` modifiers for explicit control over tag dependency semantics in `depend_on()` — `AnyOf` requires at least one tagged member to be active, `AllOf` requires all of them (bare `Tag` defaults to `AllOf`)
--- a/src/parlant/api/services.py
+++ b/src/parlant/api/services.py
@@ -84,10 +84,9 @@ class SDKServiceParamsDTO(
 ServiceOpenAPIParamsSourceField: TypeAlias = Annotated[
     str,
     Field(
-        description="""URL or filesystem path to the OpenAPI specification.
-        For URLs, must be publicly accessible.
-        For filesystem paths, the server must have read permissions.""",
-        examples=["https://api.example.com/openapi.json", "/etc/parlant/specs/example-api.yaml"],
+        description="""URL of the OpenAPI specification (http:// or https://).
+        The URL must be reachable from the server.""",
+        examples=["https://api.example.com/openapi.json"],
     ),
 ]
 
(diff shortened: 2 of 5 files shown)


andesyteoss:
Closing this as inactive — no maintainer response after 14 days.

The security finding and fix remain valid. If this is still relevant, I'm happy to reopen, rebase, or re-submit against a different branch. Just drop a comment.

PR #798, opened 2026-05-18
fix: guard against empty/filtered LLM responses in all NLP adapters

qizwiz:
## What

Add a null check before `choices[0].message` in all 12 OpenAI-compatible NLP service adapters.

**Files changed** (all in `src/parlant/adapters/nlp/`):
`azure_service.py`, `cerebras_service.py`, `deepseek_service.py`, `fireworks_service.py`, `litellm_service.py`, `mistral_service.py`, `novita_service.py`, `openai_service.py`, `openrouter_service.py`, `qwen_service.py`, `together_service.py`, `zhipu_service.py`

## Why

Two distinct crash vectors exist at `response.choices[0].message.content`:

1. **`IndexError`** — when `choices` is an empty list (network interruption, provider-side rate limit edge cases, some API-level errors that still return HTTP 200 with an empty body)
2. **`AttributeError`** — when `choices[0].message` is `None`. This is observed on Gemini via OpenAI-compatible endpoints when content is filtered (PROHIBITED_CONTENT), which returns HTTP 200 with `message: null` rather than an HTTP error code

The existing `or "{}"` guard only handles `content` being `None` or empty-string — it does not protect against either crash vector above.

## Fix

```python
if not response.choices or response.choices[0].message is None:
    raise ValueError("LLM returned empty or filtered response")
raw_content = response.choices[0].message.content or "{}"
```

This converts both silent crashes into a clear, actionable error message that callers can catch and handle appropriately.

## No behaviour change for well-formed responses

For all responses where `choices` is non-empty and `message` is not None, the code is identical to before.


mc-dorzo:
Hey man, can you please make your PR from and to the `develop` branch?

qizwiz:
Closing — this PR was mistakenly filed against `main` instead of `develop`. Refiling now against the correct base branch.

